PR #2, opened 2023-10-26
add new crawl_strategy: `pagination_only`

BurnzZ:
**Motivation**: If users only want to paginate the products through a given navigation/category page, then setting `crawl_strategy=pagination_only` become an attractive option since it removes the risk of misidentified subCategory links by ML-extraction.

### DIFF
--- a/zyte_spider_templates/spiders/ecommerce.py
+++ b/zyte_spider_templates/spiders/ecommerce.py
@@ -41,7 +41,10 @@ class EcommerceSpiderParams(BaseSpiderParams):
                 },
                 EcommerceCrawlStrategy.pagination_only: {
                     "title": "Pagination Only",
-                    "description": "Follow pagination and product detail pages.",
+                    "description": (
+                        "Follow pagination and product detail pages. SubCategory links are ignored. "
+                        "Use this when some subCategory links are misidentified by ML-extraction."
+                    ),
                 },
             },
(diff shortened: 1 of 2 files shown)


zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -30,13 +31,17 @@ class EcommerceSpiderParams(BaseSpiderParams):
...
                 EcommerceCrawlStrategy.navigation: {
                     "title": "Navigation",
                     "description": "Follow pagination, subcategories, and product detail pages.",
                 },
-                EcommerceCrawlStrategy.full: {
-                    "title": "Full",
-                    "description": "Follow most links within the domain of URL in an attempt to discover and extract as many products as possible.",
+                EcommerceCrawlStrategy.pagination_only: {
+                    "title": "Pagination Only",
+                    "description": "Follow pagination and product detail pages.",
```
kmike, line 44: What do you think about adding some motivation to the description here, similar to the PR description?

BurnzZ: Good idea. Updated in https://github.com/zytedata/zyte-spider-templates/pull/2/commits/396283d82f2a92175698d3cbd4f0dd1012310af5

wRAR (approved):

PR #5, opened 2023-10-26
Add Python 3.12 support.

wRAR:


wRAR:
(blocked on 3.12 support in aiohttp)

BurnzZ (changes requested):
Let's update this line as well: https://github.com/zytedata/zyte-spider-templates/blob/main/.readthedocs.yml#L8

kmike (approved):

PR #9, opened 2023-10-31
Initial docs

Gallaecio:


zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -13,15 +14,29 @@
 from zyte_spider_templates.spiders.base import BaseSpider, BaseSpiderParams
 
 
+@document_enum
 class EcommerceCrawlStrategy(str, Enum):
     full: str = "full"
+    """Follow most links within the domain of URL in an attempt to discover and
+    extract as many products as possible."""
```
Gallaecio, line 21: Tried to use `Foo.bar.__doc__` in `enumMeta` to avoid duplication, but the `__doc__` is not generated at regular run time (there’s some “interactive” condition in the decorator implementation), and this way we can potentially provide more details in the reference docs than those we want in enumMeta; maybe we could even make enumMeta provide a link to the upstream reference for details where needed.

README.rst
```diff
@@ -30,24 +32,7 @@ starting point for your own projects.
...
-
-    pip install zyte-spider-templates
-
-
-Spiders
-=======
-
-* ``EcommerceSpider``: a spider that can scrape products from any e-commerce
-  website.
+* Documentation: https://zyte-spider-templates.readthedocs.io/en/latest/
```
Gallaecio, line 37: :crossed_fingers: No one steals it.

zyte_spider_templates/spiders/base.py
```diff
@@ -1,3 +1,4 @@
+import re
```
BurnzZ, line 1: I think we don't need this?

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -13,15 +14,29 @@
 from zyte_spider_templates.spiders.base import BaseSpider, BaseSpiderParams
 
 
+@document_enum
```
BurnzZ, line 17: TIL

docs/setup.rst
```diff
@@ -0,0 +1,73 @@
...
+
+-   Set :setting:`CLOSESPIDER_TIMEOUT_NO_ITEM
+    <scrapy:CLOSESPIDER_TIMEOUT_NO_ITEM>` to 600, to force the spider to stop
+    if no item has been found for 10 minutes.
+
+-   Set :setting:`SCHEDULER_DISK_QUEUE <scrapy:SCHEDULER_DISK_QUEUE>` to
+    ``"scrapy.squeues.PickleFifoDiskQueue"`` and
+    :setting:`SCHEDULER_MEMORY_QUEUE <scrapy:SCHEDULER_MEMORY_QUEUE>` to
+    ``"scrapy.squeues.FifoMemoryQueue"``, for better request scheduling
+    performance.
```
BurnzZ, line 68:
```suggestion
    ``"scrapy.squeues.FifoMemoryQueue"``, for better request scheduling.
```

I'm not so sure about the performance part as the memusage, RPM, etc are still okay. It's mostly due to how the requests with the same priorities are handled.

docs/templates/e-commerce.rst
```diff
@@ -0,0 +1,24 @@
...
+
+.. code-block:: shell
+
+    scrapy crawl e-commerce -a url="https://books.toscrape.com"
+
+Parameters
+==========
+
+.. autopydantic_model:: zyte_spider_templates.spiders.ecommerce.EcommerceSpiderParams
+    :inherited-members: BaseModel
```
BurnzZ, line 18: We need to add a `description` to https://github.com/zytedata/zyte-spider-templates/blob/main/zyte_spider_templates/spiders/ecommerce.py#L52 so we can have details on this:

image

docs/templates/e-commerce.rst
```diff
@@ -0,0 +1,24 @@
...
+
+.. autoenum:: zyte_spider_templates.spiders.ecommerce.EcommerceCrawlStrategy
+
+.. autoenum:: zyte_spider_templates.spiders.ecommerce.ExtractFrom
```
BurnzZ, line 22: We need a more descriptive text for this one. The same case with **Geolocation**.

image

docs/customization/page-objects.rst
```diff
@@ -0,0 +1,146 @@
...
+
+To change or fix how a given field is extracted, overriding the value from
+`Zyte API automatic extraction`_, create a page object class, configured to run
+on some given URLs (:func:`web_poet.handle_urls`), that defines the logic to
+extract that field. For example:
+
+.. code-block:: python
+    :caption: page_objects/books_toscrape_com.py
+
+    import re
```
BurnzZ, line 35: This is unused.

docs/setup.rst
Gallaecio: “for better request priority handling”?

BurnzZ: Yep, I think that fits much better.

docs/templates/e-commerce.rst
```diff
@@ -0,0 +1,24 @@
+.. _e-commerce:
+
+===========================================
+E-commerce spider template (``e-commerce``)
```
Gallaecio, line 4:
```suggestion
E-commerce spider template (``ecommerce``)
```

Likely affects other parts of the change.

docs/customization/spiders.rst
```diff
@@ -0,0 +1,104 @@
...
+
+For example, you can make a spider that expects a product details URL and does
+not follow navigation at all:
+
+.. code-block:: python
+
+    from typing import Iterable
+
+    from scrapy import Request
+    from zyte_spider_templates import EcommerceSpider,
```
wRAR, line 92:
```suggestion
    from zyte_spider_templates import EcommerceSpider
```

docs/templates/e-commerce.rst
wRAR: Is it just a suggested change because you aren't sure?

Gallaecio: Because I realized this later, when working on docs.zyte.com, and added it here as a note to myself to fix that as I apply the rest of the feedback here later. I did not want to let this distract myself from docs.zyte.com work, and I wanted to check also if I left other references to the spider ID elsewhere.

docs/templates/e-commerce.rst
Gallaecio: In my rendering, I do not get “An enumeration.”, but rather nothing at all. Which I think is fine given the parameter descriptions themselves at the top. No strong opinion, though.

wRAR (approved):

BurnzZ (approved):

PR #13, opened 2023-11-08
Improve forbidden domain handling

Gallaecio:
I tested it manually with a raw spider defined on a https://github.com/zytedata/zyte-spider-templates-project project with the 2 setting changes needed:

```python
from zyte_spider_templates import BaseSpider, EcommerceSpider

from pydantic import BaseModel
from scrapy import Request, Spider
from scrapy_spider_metadata import Args


class Params(BaseModel):
    pass


class TestSpider(EcommerceSpider, Args[Params]):
    name = "test"

    @classmethod
    def from_crawler(cls, crawler, *args, **kwargs):
        spider = super(BaseSpider, cls).from_crawler(crawler, *args, **kwargs)
        return spider

    def start_requests(self):
        for url in (
            # Also tested adding an extra, non-forbidden URL.
            "https://forbidden.domain.example",
        ):
            yield Request(
                url=url,
                callback=self.parse_navigation,
            )
```

Should I also figure out a way to test it? It feels like writing the test will take longer than implementing the feature took, so I am not sure whether or not it is worth it. And if we do, should we aim to test both possible close paths (`_maybe_close` calls)? Would tests that create the 2 middlewares and “play” with them work, as opposed to running an actual spider on a mock local website?

### DIFF
--- a/zyte_spider_templates/middlewares.py
+++ b/zyte_spider_templates/middlewares.py
@@ -5,8 +5,9 @@ from datetime import datetime
 from typing import Any, Dict
 
 from scrapy import Request
-from scrapy.exceptions import ScrapyDeprecationWarning
+from scrapy.exceptions import CloseSpider, ScrapyDeprecationWarning
 from scrapy.utils.request import request_fingerprint
+from zyte_api.aio.errors import RequestError
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


zyte_spider_templates/middlewares.py
```diff
@@ -109,3 +110,72 @@ def crawl_logs(self, response, result):
...
+    def process_start_requests(self, start_requests, spider):
+        count = 0
+        for request in start_requests:
+            request.meta["is_start_request"] = True
+            yield request
+            count += 1
+        self._send_signal(_start_requests_processed, count=count)
+
+
+class ForbiddenDomainDownloaderMiddleware:
```
kmike, line 139: What do you think about moving this logic to scrapy-zyte-api?

PR #14, opened 2023-11-10
Allow crawling of items outside of domain

BurnzZ:
TODO: 

- [ ] Update https://github.com/zytedata/zyte-spider-templates-project to include the middleware upon release

### DIFF
--- a/docs/setup.rst
+++ b/docs/setup.rst
@@ -70,6 +70,11 @@ The following additional settings are recommended:
     ``"zyte_crawlers.middlewares.CrawlingLogsMiddleware": 1000``, to log crawl
     data in JSON format for debugging purposes.
 
+-   Update :setting:`SPIDER_MIDDLEWARES <scrapy:SPIDER_MIDDLEWARES>` to include
+    ``"zyte_crawlers.middlewares.ItemOffsiteMiddleware": 500`` and remove
+    ``"scrapy.spidermiddlewares.offsite.OffsiteMiddleware"``. This allows for the
+    ``allow_items_outside_domains`` spider parameter to work.
+
 For an example of a properly configured ``settings.py`` file, see `the one
 in zyte-spider-templates-project`_.
 
--- a/docs/setup.rst
+++ b/docs/setup.rst
@@ -71,9 +71,9 @@ The following additional settings are recommended:
     data in JSON format for debugging purposes.
 
 -   Update :setting:`SPIDER_MIDDLEWARES <scrapy:SPIDER_MIDDLEWARES>` to include
-    ``"zyte_crawlers.middlewares.ItemOffsiteMiddleware": 500`` and remove
-    ``"scrapy.spidermiddlewares.offsite.OffsiteMiddleware"``. This allows for the
-    ``allow_items_outside_domains`` spider parameter to work.
+    ``"zyte_crawlers.middlewares.AllowOffsiteMiddleware": 500`` and remove
+    ``"scrapy.spidermiddlewares.offsite.OffsiteMiddleware"``. This allows for
+    crawling item links outside of the domain.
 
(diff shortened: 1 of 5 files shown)


docs/setup.rst
```diff
@@ -70,6 +70,11 @@ The following additional settings are recommended:
     ``"zyte_crawlers.middlewares.CrawlingLogsMiddleware": 1000``, to log crawl
     data in JSON format for debugging purposes.
 
+-   Update :setting:`SPIDER_MIDDLEWARES <scrapy:SPIDER_MIDDLEWARES>` to include
+    ``"zyte_crawlers.middlewares.ItemOffsiteMiddleware": 500`` and remove
+    ``"scrapy.spidermiddlewares.offsite.OffsiteMiddleware"``. This allows for the
```
BurnzZ, line 75: What do you think about moving this to Scrapy?

zyte_spider_templates/spiders/base.py
```diff
@@ -49,6 +49,10 @@ class BaseSpiderParams(BaseModel):
             "widget": "request-limit",
         },
     )
+    allow_items_outside_domains: Optional[bool] = Field(
+        description="When set to True, items outside of the domains will be crawled.",
+        default=False,
```
kmike, line 54: Do you think it's feasible NOT to add this option, but make it a default instead? It looks safe enough, at least for "navigation" and "pagination_only" strategies. Maybe it's fine for "full" as well. We might need to do some experiment to ensure it's safe though.

BurnzZ: After some analysis, it turns out we can enabled crawling products outside of the domain by default. I guess the ultimate switch by users here would be enabling the `AllowOffsiteMiddleware` in settings.

docs/setup.rst
BurnzZ: Created a proposal for Scrapy in https://github.com/scrapy/scrapy/pull/6151

kmike (approved):

docs/setup.rst
```diff
@@ -70,6 +70,11 @@ The following additional settings are recommended:
     ``"zyte_crawlers.middlewares.CrawlingLogsMiddleware": 1000``, to log crawl
     data in JSON format for debugging purposes.
 
+-   Update :setting:`SPIDER_MIDDLEWARES <scrapy:SPIDER_MIDDLEWARES>` to include
+    ``"zyte_crawlers.middlewares.AllowOffsiteMiddleware": 500`` and remove
+    ``"scrapy.spidermiddlewares.offsite.OffsiteMiddleware"``. This allows for
+    crawling item links outside of the domain.
```
Gallaecio, line 76: This being a default middleware, it needs to be disabled, not “removed”, which might not be clear for some users. Maybe…

```suggestion
-   Update :setting:`SPIDER_MIDDLEWARES <scrapy:SPIDER_MIDDLEWARES>` to include
    ``"zyte_spider_templates.middlewares.AllowOffsiteMiddleware": 500`` and
    ``"scrapy.spidermiddlewares.offsite.OffsiteMiddleware": None``. This allows for
    crawling item links outside of the domain.
```

zyte_spider_templates/middlewares.py
```diff
@@ -109,3 +110,13 @@ def crawl_logs(self, response, result):
...
+
+
+class AllowOffsiteMiddleware(OffsiteMiddleware):
+    def _filter(self, request: Any, spider: Spider) -> bool:
+        if not isinstance(request, Request):
+            return True
+        if request.meta.get("allow_offsite"):
+            return True
+        result = super()._filter(request, spider)
+        return result
```
Gallaecio, line 122: :lipstick: 

```suggestion
        return super()._filter(request, spider)
```

Gallaecio (approved):

PR #19, opened 2023-12-14
page_objects → pages

Gallaecio:
To-do:
- [x] Maintain backward compatibility.


kmike (approved):

wRAR (approved):

BurnzZ (approved):

Gallaecio:
Sorry to bother you again, but please review https://github.com/zytedata/zyte-spider-templates/pull/19/commits/e811fa4274fc0aad920e59d94b5a2534213f4e45 as well, just to be sure before I merge.

tests/pages/test_product_navigation_heuristics.py
```diff
@@ -125,3 +126,16 @@ async def test_crawl_nofollow_links():
...
+    with pytest.warns(DeprecationWarning, match="page_objects"):
+        from zyte_spider_templates.page_objects import (  # noqa: F401
+            HeuristicsProductNavigationPage,
+        )
+
+    # We cannot test the warning again because duplicate warnings are ignored,
+    # but we still want to ensure that we can import the class.
+    from zyte_spider_templates.page_objects.product_navigation_heuristics import (  # noqa: F401, F811
+        HeuristicsProductNavigationPage,
+    )
```
BurnzZ, line 141: If it's the successful import we want to check, what do you think about moving this 2nd check to another test function? so that we can assert if a warning was generated or not.

kmike (approved):

tests/pages/test_product_navigation_heuristics.py
Gallaecio: I tried that, does not help :cry:

BurnzZ: Fair enough :D 

I'm not sure if https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.deprecated_call would also help here.

BurnzZ (approved):

tests/pages/test_product_navigation_heuristics.py
Gallaecio: Also tried that :sweat:

I think it is Python itself that ignores duplicate warnings. There might be a way to do it, but I could not find it in a reasonable time, and I am not sure if it is worth the effort here.

PR #21, opened 2023-12-18
log to stats when items are dropped due to low proba

BurnzZ:
Need this to easily check experiment runs and compare how this stat changes.

### DIFF
--- a/zyte_spider_templates/spiders/ecommerce.py
+++ b/zyte_spider_templates/spiders/ecommerce.py
@@ -171,6 +171,7 @@ class EcommerceSpider(Args[EcommerceSpiderParams], BaseSpider):
         if probability is None or probability >= 0.1:
             yield product
         else:
+            self.crawler.stats.inc_value("drop_item/product/low_probability")
             self.logger.info(
                 f"Ignoring item from {response.url} since its probability is "
                 f"less than threshold of 0.1:\n{product}"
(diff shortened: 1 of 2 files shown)


Gallaecio (approved):

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -171,6 +171,7 @@ def parse_product(
         if probability is None or probability >= 0.1:
             yield product
         else:
+            self.crawler.stats.inc_value("drop_item/product/low_probability")
```
PyExplorer, line 174: Just wondering if we have any agreement for stats naming at this time? Should it start from product, for example, like product/drop_item/low_probability for easy tracking in the future when we have several stats for product?

Gallaecio: No strong opinion, although I think the proposed naming makes sense. I do wonder if we should have some more prefix (e.g. `spider/ecommerce/`), but no strong opinion on that either.

BurnzZ: No strong opinion in the naming for now as patterns will emerge later on.

PyExplorer: yes, agree, thanks, let's back to this later on.

PyExplorer (approved):

PR #22, opened 2023-12-22
avoid crawling nextPage if no product links

BurnzZ:
This avoids the False Positive results from productNavigation pages where in reality, it's not a productNavigation page but the nextPage has been extracted. Having no product links at all is a good indicator of this False Positivity.

This change should reduce the amount of unnecessary nextPage requests from the spider.


Gallaecio (approved):

PyExplorer (approved):

kmike:
Nice heuristic!

PR #26, opened 2024-01-10
moved eCommerce specific code from BaseSpider

BurnzZ:
Copy and Pasted code.

This is in preparation of having the ArticleSpider soon.


zyte_spider_templates/settings.py
```diff
@@ -0,0 +1,42 @@
+from itemadapter import ItemAdapter
```
kmike, line 1: What's the use case for this file?

BurnzZ: Good catch! This should not have been included as it's used locally for testing.

kmike (approved):

PR #27, opened 2024-01-12
remove 'www*.' prefix when parsing allowed_domains

BurnzZ:
This fixes this specific use case:
* User has entered the parameter: `url=https://www.example.com`.
* This sets `allowed_domains=["www.example.com"]`.
* However, the site lists its links as `https://example.com/some-page` _(without the 'www.' prefix)_
* This causes the `OffsiteMiddleware` to filter out the links.

Removing any 'www*.' prefixes beforehand allows for a cleaner **allowed_domains** value.

The reverse should also work where `url=https://example.com` is used as the arg which sets `allowed_domains=["example.com"]` and the site uses links like `https://www.example.com/some-page`. This is because, the `allowed_domain` would always be a subset of the links in the `OffsiteMiddleware` ([code ref](https://github.com/scrapy/scrapy/blob/master/scrapy/spidermiddlewares/offsite.py#L67-L71)).

Lastly, also moved parsing the **allowed_domains** from the BaseSpider to the EcommerceSpider since ArticleSpider would have a different way to parse them due to multiple seeds.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,0 +1,14 @@
+import pytest
+
+from zyte_spider_templates.utils import get_domain
+
+URL_TO_DOMAIN = (
+    ("https://example.com", "example.com"),
+    ("https://www.example.com", "example.com"),
+    ("https://www2.example.com", "example.com"),
+)
+
+
+@pytest.mark.parametrize("url,domain", URL_TO_DOMAIN)
+def test_get_domain(url, domain):
+    assert get_domain(url) == domain
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -6,6 +6,12 @@ URL_TO_DOMAIN = (
     ("https://example.com", "example.com"),
     ("https://www.example.com", "example.com"),
     ("https://www2.example.com", "example.com"),
+    ("https://prefixwww.example.com", "prefixwww.example.com"),
+    ("https://wwworld.example.com", "wwworld.example.com"),
+    ("https://my.wwworld-example.com", "my.wwworld-example.com"),
+    ("https://wwwow.com", "wwwow.com"),
+    ("https://wowww.com", "wowww.com"),
+    ("https://awww.com", "awww.com"),
 )
 
 
--- a/zyte_spider_templates/utils.py
+++ b/zyte_spider_templates/utils.py
@@ -4,6 +4,4 @@ from scrapy.utils.url import parse_url
 
 
 def get_domain(url: str) -> str:
-    allowed_domain = parse_url(url).netloc
-    allowed_domain = re.sub(r"www.*?\.", "", allowed_domain)
-    return allowed_domain
+    return re.sub(r"www.*?\.", "", parse_url(url).netloc)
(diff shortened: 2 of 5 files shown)


Gallaecio (approved):
Looks good to me.

Although I wonder if a better approach might be to find the top-level domain part of the URL (with `tldextract`, which Scrapy uses), and include that plus the subdomain. So, `a.b.c.example.com` becomes `example.com`. It’s just a thought, though, and I am not even sure if it would be a change for the better or for the worse.

BurnzZ:
> Although I wonder if a better approach might be to find the top-level domain part of the URL (with tldextract, which Scrapy uses), and include that plus the subdomain. So, a.b.c.example.com becomes example.com. It’s just a thought, though, and I am not even sure if it would be a change for the better or for the worse.

This might be the case later on as @PyExplorer polishes the Article Spider which has quite lots of varying cases with the domains.

zyte_spider_templates/utils.py
```diff
@@ -0,0 +1,9 @@
+import re
+
+from scrapy.utils.url import parse_url
+
+
+def get_domain(url: str) -> str:
+    allowed_domain = parse_url(url).netloc
```
wRAR, line 7: Can you please rename the var as the code doesn't look specific to allowed domains?

BurnzZ: Good call. https://github.com/zytedata/zyte-spider-templates/pull/27/commits/c66f4ddb9a59577f6046e8ab27375c5e0ac9f4f8

wRAR (approved):

PyExplorer (approved):

tests/test_utils.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+
+from zyte_spider_templates.utils import get_domain
+
+URL_TO_DOMAIN = (
+    ("https://example.com", "example.com"),
+    ("https://www.example.com", "example.com"),
+    ("https://www2.example.com", "example.com"),
```
kmike, line 8: Could you please add a test where url contains www., but doesn't start with it?

zyte_spider_templates/utils.py
```diff
@@ -0,0 +1,7 @@
+import re
+
+from scrapy.utils.url import parse_url
+
+
+def get_domain(url: str) -> str:
+    return re.sub(r"www.*?\.", "", parse_url(url).netloc)
```
kmike, line 7: Would it make sense to have this function more conservative?  It looks like this one can change netlocs like `wwworld.example.com`, or `my.wwworld.com`, or `awww.com`. 

What are the real-world cases besides www.example.com which we should support?

kmike (approved):

zyte_spider_templates/utils.py
```diff
@@ -4,4 +4,4 @@
 
 
 def get_domain(url: str) -> str:
-    return re.sub(r"www.*?\.", "", parse_url(url).netloc)
+    return re.sub(r"^(www)\d?\.", "", parse_url(url).netloc)
```
Gallaecio, line 7:
```suggestion
    return re.sub(r"^www\d*\.", "", parse_url(url).netloc)
```

The parenthesis were unnecessary, and `*` works for www54.example.com as well (if we want that).

BurnzZ: Good call!

Issue #25, opened 2024-01-10
Discussion: Handle data type of `response` data type in HeuristicsProductNavigationPage (`BrowserResponse` vs `HttpResponse`)


BurnzZ:
# Problem

Currently, `HeuristicsProductNavigationPage` uses `HttpResponse` as one of the dependencies ([code ref](https://github.com/zytedata/zyte-spider-templates/blob/main/zyte_spider_templates/pages/product_navigation_heuristics.py#L17-L25)). This causes the spider to send 2 separate requests to ZAPI if the `extract_from` parameter is set to "browserHtml":
- `{"productNavigationOptions": {"extractFrom": "browserHtml"}, "productNavigation": true, "url": "https://books.toscrape.com/"}`
- `{"httpResponseBody": true, "url": "https://books.toscrape.com/"}`

This is also true when it's left to the default value which is `extract_from=None`.

Note that this issue is not related to https://github.com/scrapy-plugins/scrapy-zyte-api/issues/91 since https://github.com/scrapinghub/scrapy-poet/pull/175 has addressed it. 

The problem that we face is that the dependency of `HeuristicsProductNavigationPage` should be dynamic depending on the `extract_from` input value:
- `response: BrowserResponse`, if `extract_from=browserHtml`
- `response: HttpResponse`, if `extract_from=httpResponseBody`

kmike:
I think https://github.com/scrapinghub/andi/pull/30 should allow to express this kind of dependencies in page objects.

BurnzZ:
Yes, it would 👍 

Though some new interface in page objects should be made to allow declaration of dynamic dependencies. For example, depending on the `extract_from` value, it would logically switch between `BrowserResponse` or `HttpResponse`.

PyExplorer:
@BurnzZ I've checked this approach with `AnyResponse` for Articles and it's actually resolves handling the case with extract_from (browserHtml or httpResponseBody) argument for spider. 
But the issue with doubling the requests is still here.

PyExplorer:
After double checking I see that this is working, thanks! 
In my case, previously, doubled requests were only for HttpResponse, and for BrowserResponse everything was in one response.
[...]

PR #28, opened 2024-01-16
support AnyResponse

BurnzZ:
Fixes https://github.com/zytedata/zyte-spider-templates/issues/25.

Related PRs:
- https://github.com/scrapinghub/web-poet/pull/195
- https://github.com/scrapinghub/scrapy-poet/pull/184
- https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161

# TODO:
- [x] Use new release of `scrapy-zyte-api`
- [x] Use new release of `scrapy-poet`

### DIFF
--- a/zyte_spider_templates/spiders/base.py
+++ b/zyte_spider_templates/spiders/base.py
@@ -29,8 +29,9 @@ class ExtractFrom(str, Enum):
 class BaseSpiderParams(BaseModel):
     url: str = Field(
         title="URL",
-        description="Initial URL for the crawl.",
-        pattern=r"^https?:\/\/[^:\/\s]+(:\d{1,5})?(\/[^\s]*)*(#[^\s]*)?$",
+        description="Initial URL for the crawl. Enter the full URL including http(s), "
+        "you can copy and paste it from your browser. Example: https://toscrape.com/",
+        pattern=r"^https?://[^:/\s]+(:\d{1,5})?(/[^\s]*)*(#[^\s]*)?$",
     )
     geolocation: Optional[Geolocation] = Field(
         title="Geolocation",
--- a/zyte_spider_templates/spiders/base.py
+++ b/zyte_spider_templates/spiders/base.py
@@ -1,3 +1,4 @@
+from enum import Enum
 from importlib.metadata import version
 from typing import Any, Dict, Optional
 
@@ -9,9 +10,20 @@ from zyte_spider_templates._geolocations import (
     GEOLOCATION_OPTIONS_WITH_CODE,
     Geolocation,
 )
+from zyte_spider_templates.documentation import document_enum
 
 # Higher priority than command-line-defined settings (40).
 ARG_SETTING_PRIORITY: int = 50
+
+
+@document_enum
+class ExtractFrom(str, Enum):
+    httpResponseBody: str = "httpResponseBody"
+    """Use HTTP responses. Cost-efficient and fast extraction method, which
+    works well on many websites."""
+
+    browserHtml: str = "browserHtml"
+    """Use browser rendering. Often provides the best quality."""
 
 
 class BaseSpiderParams(BaseModel):
@@ -45,6 +57,26 @@ class BaseSpiderParams(BaseModel):
         default=100,
         json_schema_extra={
             "widget": "request-limit",
+        },
+    )
+    extract_from: Optional[ExtractFrom] = Field(
+        title="Extraction source",
+        description=(
+            "Whether to perform extraction using a browser request "
+            "(browserHtml) or an HTTP request (httpResponseBody)."
+        ),
+        default=ExtractFrom.browserHtml,
+        json_schema_extra={
+            "enumMeta": {
+                ExtractFrom.browserHtml: {
+                    "title": "browserHtml",
+                    "description": "Use browser rendering. Often provides the best quality.",
+                },
+                ExtractFrom.httpResponseBody: {
+                    "title": "httpResponseBody",
+                    "description": "Use HTTP responses. Cost-efficient and fast extraction method, which works well on many websites.",
+                },
+            },
         },
     )
 
--- a/zyte_spider_templates/spiders/base.py
+++ b/zyte_spider_templates/spiders/base.py
@@ -66,7 +66,7 @@ class BaseSpiderParams(BaseModel):
             "Whether to perform extraction using a browser request "
             "(browserHtml) or an HTTP request (httpResponseBody)."
         ),
-        default=ExtractFrom.browserHtml,
+        default=None,
         json_schema_extra={
             "enumMeta": {
                 ExtractFrom.browserHtml: {
--- a/zyte_spider_templates/spiders/ecommerce.py
+++ b/zyte_spider_templates/spiders/ecommerce.py
@@ -74,7 +74,7 @@ class EcommerceSpiderParams(BaseSpiderParams):
             "Whether to perform extraction using a browser request "
             "(browserHtml) or an HTTP request (httpResponseBody)."
         ),
-        default=None,
+        default=ExtractFrom.browserHtml,
         json_schema_extra={
             "enumMeta": {
                 ExtractFrom.browserHtml: {
--- a/zyte_spider_templates/spiders/ecommerce.py
+++ b/zyte_spider_templates/spiders/ecommerce.py
@@ -33,16 +33,6 @@ class EcommerceCrawlStrategy(str, Enum):
     ML-extraction."""
 
 
-@document_enum
-class ExtractFrom(str, Enum):
-    httpResponseBody: str = "httpResponseBody"
-    """Use HTTP responses. Cost-efficient and fast extraction method, which
-    works well on many websites."""
-
-    browserHtml: str = "browserHtml"
-    """Use browser rendering. Often provides the best quality."""
-
-
 class EcommerceSpiderParams(BaseSpiderParams):
     crawl_strategy: EcommerceCrawlStrategy = Field(
         title="Crawl strategy",
@@ -64,26 +54,6 @@ class EcommerceSpiderParams(BaseSpiderParams):
                         "Follow pagination and product detail pages. SubCategory links are ignored. "
                         "Use this when some subCategory links are misidentified by ML-extraction."
                     ),
-                },
-            },
-        },
-    )
-    extract_from: Optional[ExtractFrom] = Field(
-        title="Extraction source",
-        description=(
-            "Whether to perform extraction using a browser request "
-            "(browserHtml) or an HTTP request (httpResponseBody)."
-        ),
-        default=ExtractFrom.browserHtml,
-        json_schema_extra={
-            "enumMeta": {
-                ExtractFrom.browserHtml: {
-                    "title": "browserHtml",
-                    "description": "Use browser rendering. Often provides the best quality.",
-                },
-                ExtractFrom.httpResponseBody: {
-                    "title": "httpResponseBody",
-                    "description": "Use HTTP responses. Cost-efficient and fast extraction method, which works well on many websites.",
                 },
             },
         },
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.6.0
+current_version = 0.6.1
 commit = True
 tag = True
 tag_name = {new_version}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,8 +47,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.12"]
-        tox-job: ["mypy", "linters", "twine", "docs"]
+        include:
+        - python-version: '3.12'
+          toxenv: mypy
+        - python-version: '3.12'
+          toxenv: linters
+        - python-version: '3.12'
+          toxenv: twine
+        - python-version: '3.11'
+          toxenv: docs
 
     steps:
     - uses: actions/checkout@v4
@@ -62,4 +69,4 @@ jobs:
         python -m pip install tox
     - name: tox
       run: |
-        tox -e ${{ matrix.tox-job }}
+        tox -e ${{ matrix.toxenv }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,6 +18,8 @@ jobs:
       matrix:
         include:
         - python-version: '3.8'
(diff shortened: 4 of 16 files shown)


zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -69,7 +69,7 @@ class EcommerceSpiderParams(BaseSpiderParams):
             "Whether to perform extraction using a browser request "
             "(browserHtml) or an HTTP request (httpResponseBody)."
         ),
-        default=None,
+        default=ExtractFrom.browserHtml,
```
BurnzZ, line 72: Hardcoding this should be okay since we don't have Zyte API's Auto-Configurator yet. So somewhere in the backend, a `None` value is still hardcoded.

PyExplorer: @BurnzZ what do you think about moving this to BaseSpiderParams? Otherwise I also need to redefine `extract_from` for Articles.

BurnzZ: Sounds good to me! https://github.com/zytedata/zyte-spider-templates/pull/28/commits/8d7130fe6fd9c7984d76cc0ed040bd7ac006b4b2

zyte_spider_templates/spiders/base.py
```diff
@@ -48,6 +60,26 @@ class BaseSpiderParams(BaseModel):
             "widget": "request-limit",
         },
     )
+    extract_from: Optional[ExtractFrom] = Field(
+        title="Extraction source",
+        description=(
+            "Whether to perform extraction using a browser request "
+            "(browserHtml) or an HTTP request (httpResponseBody)."
+        ),
+        default=ExtractFrom.browserHtml,
```
kmike, line 69: Could you please elaborate, why is this change needed?

BurnzZ: Since Zyte API's Auto-Configurator isn't released yet, we can't retrieve the extraction source that was used on a given page. This forces the a default page type to be included in the Zyte API request.

We can revert back to `None` when the Auto-Configurator is released soon as we'd have a way to get either `browserHtml` or `httpResponseBody` that was used.

kmike: Is this change fixing something, or is it done just in case?

BurnzZ: It's done so that `ZyteApiProvider` can properly request the proper extraction source when building the Zyte API Request: https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161/files#diff-67cea92ffa3989fe30ffd7f4bbe868f953810e0f4fdf2ddfe7f18bc54fba505eR125-R135

kmike: How would it work if a user changes it back to None?

BurnzZ: I've reverted this change @kmike after confirming that `AnyResponse` would take `httpResponse` by default if `browserResponse` isn't available. https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161#discussion_r1458663646. 

Now that `extract_from=None` by default, it'd use `httpResponse`.

https://github.com/zytedata/zyte-spider-templates/pull/28/commits/e766bc413a0f2ed68e11314d9225b7202866a0f1

BurnzZ: @kmike  I just realized that using `httpResponse` by default is a change that might cause the customer some regressions in terms of extraction quality as `browserResponse` is being used by Zyte API backend as the default extraction source.

kmike: Yes, this is actually what triggered https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161#discussion_r1471812502 initially. 

It seems that in absence of anyResponse option in the API, when extraction is requested, we can either 

1. default to extractFrom=httpResponseBody,
2. default to extractFrom=browserHtml,
3. not set extractFrom, and request httpResponseBody
4. not set extractFrom, and request browserHtml.

I assume (3) doesn't work, and (4) is the same is (2) now, so options to consider are (1) and (2), but I haven't checked it.

Let's consider the implications.

First, with the current state (2) looks like clearly the way to go. It matches what's going on in the API, and if user wants to avoid browser, they can configure it in the spider parameters.

Things get more complicated if the API starts to pick render vs non-render automatically, with most websites using non-render option. Switching to render when the API uses raw could have unexpected price regressions (though I'm not sure), and quality regressions as well. Switching to httpResponseBody when the API is using rendering behind the scenes is also not great, as it's actually is going to be almost guaranteed to cause quality regression (backend is only going to be used render when it's needed to get the quality). So, it seems there is no a good automatic solution in this case, without introducting the anyResponse parameter.

Let's say the automatic system is in place, but anyResponse option is not there in API. We have 2 options: keep rendered default, or make a switch to httpResponseBody as a default. If we keep the rendered default, then the users of smart spiders won't be benefitting from the auto-switch feature. If we switch to httpResponseBody as a default, on some websites there is going to be a regression.

Either way, it seems the "default=null" is not going to work as expected if AnyResponse is requested, unless there is anyResponse API argument. Instead, it does the same as one of the other options: extractFrom=httpResponseBody or extractFrom=browserHtml. We may just as well remove the `null` support from the spider, though there could be cases the option still works, if a strategy doesn't use AnyResponse.

BurnzZ: Thanks for laying it out @kmike . It would seem that the best course of action would be:

1. Set `browserHtml` as the default `extract_from` parameter.
2. Remove `None` as one of the `extract_from` parameter.
3. When Zyte API Auto-Configurator is already in-place _(i.e. we can set `anyResponse=True` or something similar)_, we can:
    * bring back `None` as the one of the `extract_from` parameters.
    * set `None` as the default.

(1) and (2) ensures that there's no regressions with the current way of crawling.

(3) enables compatibility with the Zyte API Auto-Configurator. Although there will be regressions _(e.g. some websites are now extracting from browserHtml to HttpResponse, vise-versa)_ but it's coming from Zyte API and would be unavoidable.

Is this close to what you're thinking?

Gallaecio: Should we consider failing instead of making a choice? i.e. We raise an exception with a message indicating that, in this scenario, the user must specify the extractFrom parameter in ZYTE_API_PROVIDER_PARAMS.

kmike: It seems the "ideal" scenario is the following:

1. For now, we use extractFrom=browserHtml when user doesn't state explicitly the extraction method, and AnyResponse is needed by a strategy. This behavior is consistent with what the API is doing, and there are no regressions here. We keep the default=null option. 
2. Then, API is released with automatic config AND anyResponse flag. We switch the default=null to use this flag when appropriate, instead of hardcoding extractFrom=browserHtml. Users with default=null option get faster&cheaper crawls.

The issue happens if automatic config is released without anyResponse flag.

BurnzZ: Thanks @kmike , updated to have `default=ExtractFrom.browserHtml`.

BurnzZ: > For now, we use extractFrom=browserHtml when user doesn't state explicitly the extraction method, and AnyResponse is needed by a strategy. 

@kmike to clarify, shall we change the behavior directly from scrapy-zyte-api? which _diverges_ from the idea that when neither BrowserResponse nor HttpResponse is available, the default would be HttpResponse. Ref: https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161#discussion_r1458700170

or that we set extractFrom=browserHtml here as the default so it'd be used when the spider doesn't explicitly state the extraction method?

kmike: I was thinking about a change in scrapy-zyte-api, yes. 

> which diverges from the idea that when neither BrowserResponse nor HttpResponse is available, the default would be HttpResponse.

I think that's exactly the issue we're having here: defaulting to HttpResponse in scrapy-zyte-api can now cause a regression in extraction quality for scrapy-zyte-api users, while defaulting to BrowserHtml doesn't have any negative effect right now. 

Things change when the API implements auto-switching; at this point a dedicated anyReponse API parameter seems like the way to go, but there is a chance we won't have it.

kmike: About the defaults: scrapy-zyte-api can still use HttpResponse as a default when there are no other parameters which require browser rendering.

BurnzZ: > scrapy-zyte-api can still use HttpResponse as a default when there are no other parameters which require browser rendering.

I'm a bit lost since this would still make the issue persist with a regression from browserResponse to httpResponse as the default. I might be missing something here.

> For now, we use extractFrom=browserHtml when user doesn't state explicitly the extraction method, and AnyResponse is needed by a strategy. This behavior is consistent with what the API is doing, and there are no regressions here. We keep the default=null option.

Perhaps I misunderstood that this one should be done inside zyte-spider-templates instead, something like:

```python
@classmethod
def from_crawler(cls, crawler: Crawler, *args, **kwargs) -> scrapy.Spider:
    spider = super(EcommerceSpider, cls).from_crawler(crawler, *args, **kwargs)
    spider.allowed_domains = [get_domain(spider.args.url)]

    extract_from = spider.args.extract_from
    if extract_from is None:
        extract_from = "browserHtml"
        spider.logger.warning(
            "The default extract_from=None is transformed into "
            "extract_from=browserHtml for now. This will change "
            "later on Zyte API can dynamically support which is "
            "the best extract source to use for a give website."
        )

    spider.settings.set(
        "ZYTE_API_PROVIDER_PARAMS",
        {
            "productOptions": {"extractFrom": extract_from},
            "productNavigationOptions": {
                "extractFrom": extract_from
            },
            **spider.settings.get("ZYTE_API_PROVIDER_PARAMS", {}),
        },
        priority=ARG_SETTING_PRIORITY,
    )

    return spider
```

kmike: > I'm a bit lost since this would still make the issue persist with a regression from browserResponse to httpResponse as the default. I might be missing something here.

Hm, why would there be a regression, from scrapy-zyte-api point of view? I assume that the code which uses AnyResponse should work fine both with HttpResponse and BrowserResponse, that's why it's declaring AnyResponse dependency. So, when some code declares that it needs AnyResponse, and this code gets HttpResponse, we don't consider it as a regression - it's the expected behavior.

If there is any other parameter which requires rendering (e.g. browserHtml: true, screenshot: true, or product: true in the current state), it makes more sense to switch to BrowserResponse - there is little extra overhead in this case, and there could be issues with trying to get HttpResponse.

If there are no other parameters, or no other parameters which require rendering, we can default to HttpResponse - it shouldn't affect anything else, just make AnyResponse use a cheaper method.

BurnzZ: > Hm, why would there be a regression, from scrapy-zyte-api point of view? 

Sorry, I was referring to zyte-spider-templates.

> I assume that the code which uses AnyResponse should work fine both with HttpResponse and BrowserResponse, that's why it's declaring AnyResponse dependency. So, when some code declares that it needs AnyResponse, and this code gets HttpResponse, we don't consider it as a regression - it's the expected behavior.

This is currently the case with scrapy-zyte-api and follows the expected behavior. 👍 

> If there are no other parameters, or no other parameters which require rendering, we can default to HttpResponse - it shouldn't affect anything else, just make AnyResponse use a cheaper method.

I guess my confusion comes from the fact that the ecommerce spider, uses HeuristicsProductNavigationPage  which have the following dependencies:

- `ProductNavigation`
- `AnyResponse`
- `PageParams`

In this case, there are no other parameters which require explicit rendering and so scrapy-zyte-api requests for HttpResponse.

kmike: > In this case, there are no other parameters which require explicit rendering and so scrapy-zyte-api requests for HttpResponse.

But currently ProductNavigation does require rendering, right?

BurnzZ: Currently, if `ProductNavigation` is requested alongside `AnyResponse`, then it would request the following in Zyte API:

```python
{
    "url": url,
    "productNavigation": True,
    "productNavigationOptions": {"extractFrom": "httpResponseBody"},
    "httpResponseBody": True,
    "httpResponseHeaders": True,
}
```

Okay hmmm, I guess I already understand what you mean from the previous comment: _"1. For now, we use extractFrom=browserHtml when user doesn't state explicitly the extraction method, and AnyResponse is needed by a strategy."_

So in this case, having the `ProductNavigation` and `AnyResponse` parameters should have:

```python
{
    "url": url,
    "producNavigation": True,
    "productOptions": {"extractFrom": "browserHtml"},
    "browserHtml": True,
}
```

But if it's only `AnyResponse`, then it'd be `httpResponseBody` and `httpResponseHeaders` that are requested.

kmike: Or maybe even

```
{
    "url": url,
    "producNavigation": True,
    "browserHtml": True,
}
```

BurnzZ: Thanks for all the clarifications @kmike ! 

Updated scrapy-zyte-api https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161/commits/a51f96173e82fcead7e30578cb1ee78ecc67cfdf and zyte-spider-templates https://github.com/zytedata/zyte-spider-templates/pull/28/commits/53b9dfd349291d08253665ff04e204a6f929b39e according to this.

BurnzZ:
Failing `docs` CI seems to be due to a cache issue in the rst files. I cannot reproduce it locally.

Gallaecio:
I cannot reproduce it either, but there should be no caching in CI, so I am quite puzzled here.

BurnzZ:
One of the great mysteries of GH 😄 I presume it would resolve itself on master.

Gallaecio:
Solved the mystery! Merge the main branch. (CI does not actually run in this branch, but in a virtual merge with the main one)

BurnzZ:
No luck there. I'm tempted to just create another PR once new versions of scrapy-poet and scrapy-zyte-api have been released. It'd be cleaner all together.

Gallaecio:
I can actually reproduce the issue after merging the main branch locally:

```
git fetch zytedata  # Yours is probably origin
git merge zytedata/main
tox -e docs
```

BurnzZ:
Aha, I see the issue now. Thanks for the help @Gallaecio ! 🙌

Gallaecio (approved):

wRAR (approved):

PyExplorer (approved):

PR #34, opened 2024-02-06
Define standard params

Gallaecio:
To do:
- [ ] Agree on the approach.
    - [ ] How to we handle `CrawlStrategy`? Do we define different “AllParams” per vertical?
- [ ] Complete, merge and release https://github.com/scrapy-plugins/scrapy-spider-metadata/pull/11
- [ ] Tests.
- [ ] Documentation.
- [ ] MyPy.

### DIFF
--- a/zyte_spider_templates/params.py
+++ b/zyte_spider_templates/params.py
@@ -1,0 +1,147 @@
+from enum import Enum
+from typing import Optional
+
+from pydantic import BaseModel, ConfigDict, Field, create_model
+from pydantic.fields import PydanticUndefined
+
+from zyte_spider_templates._geolocations import (
+    GEOLOCATION_OPTIONS_WITH_CODE,
+    Geolocation,
+)
+from zyte_spider_templates.documentation import document_enum
+
+
+@document_enum
+class CrawlStrategy(str, Enum):
+    full: str = "full"
+    """Follow most links within the domain of URL in an attempt to discover and
+    extract as many products as possible."""
+
+    navigation: str = "navigation"
+    """Follow pagination, subcategories, and product detail pages."""
+
+    pagination_only: str = "pagination_only"
+    """Follow pagination and product detail pages. SubCategory links are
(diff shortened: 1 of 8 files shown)


wRAR:
This now has conflicts.

zyte_spider_templates/params.py
```diff
@@ -0,0 +1,147 @@
...
+                pass
+        fields[param] = (field.annotation, field)
+    model = create_model(
+        cls_name,
+        __config__=ConfigDict(extra="forbid"),
+        **fields,
+    )
+    if set_args:
+        model.set_args = set_args
+    return model
```
Gallaecio, line 147: Before further progress, I would like to decide if this is the way we want to approach this, i.e. a single model with all param definitions and a function to create a model with a subset of that.

Gallaecio:
@kmike Regarding taking field reuse further, I tried removing the annotation of `extract_source` from the models into the `Field` instance (as `annotation`) in https://github.com/zytedata/zyte-spider-templates/pull/38, but it still triggers: `Field 'extract_from' requires a type annotation`.

Looking at [their error handling code](https://github.com/pydantic/pydantic/blob/12ebdfc6790ab0c29cc8aefd1d97dd04603eb7cb/pydantic/_internal/_model_construction.py#L374-L390), I think using a type annotation is a hard requirement.

Gallaecio:
@kmike Regarding the option of subclassing the `AllParams` class and hiding fields instead of the `make_params` option, it seems technically doable: we can redefine fields, and we can make it so that they are excluded from the schema, it seems.

However, I am not convinced about the approach, i.e. forcing every subclass to declare every field they want to hide, as opposed to defining the fields they want to have. I imagine with time the number of hidden fields would be larger than the number of kept fields. And when a new (optional) field is added upstream, it will show up in the UI after updating the library until hidden through code.

Gallaecio:
Closing in favor of https://github.com/zytedata/zyte-spider-templates/pull/46.

Issue #35, opened 2024-02-08
Allow extraction of Item data without discovery process


VMRuiz:
### **Description:**
Currently, the ecommerce spider in the repository accepts an input URL for an ecommerce website as input and then performs a discovery process to find Products. There are some use cases where it is required to directly input a Product URL to retrieve specific product information without the need for website crawling. _For example, for monitoring stock or price changes in a particular product, or to decouple the discovery and extraction processes._

### **Proposed Solution:**

1. **Update Ecommerce spider**: Add a new crawling strategy `EcommerceCrawlStrategy` named `disabled` to bypass any further navigation on the provided page. The spider should return the expected output Item type, `Product` in this case.

3. **Create a new Spider**: Instead of extending the current existing spider, we can build a new spider that is only focused on extraction, without any crawling functionality. In this case, we could drop the `crawl_strategy` and `max_requests` input parameters. Additionally, we can make the spider work with other item types a part from Product by adding an Output Type Selector with values like: Product, ProductNavigation, Article, etc...

I prefer the second solution because it's more versatile and keeps the alignment of their original concept on the ecommerce spider.

Gallaecio:
Option 2 sounds good to me.

Gallaecio:
We could call it `ProductSpider`, and it may be an argument for calling the upcoming article spider `ArticlesSpider`, to leave room for an `ArticleSpider`.

VMRuiz:
Does it make sense to have one different spider per item type? We may end up with tens of different templates just for this.

BurnzZ:
> There are some use cases where it is required to directly input a Product URL to retrieve specific product information without the need for website crawling. 

What would be the main advantage of using zyte-spider-templates compared to simply using Zyte API directly to perform product extraction? example: https://docs.zyte.com/zyte-api/usage/extract.html

or is this more of like leveraging the features in Scrapy Cloud like periodic jobs, logging, storing items in Hubstorage, etc?

Gallaecio:
It can be helpful for a no-code path, specially when in combination with #36.

PR #39, opened 2024-02-15
Implement crawl_strategy=direct_product

Gallaecio:
Resolves https://github.com/zytedata/zyte-spider-templates/issues/35


zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -32,6 +32,10 @@ class EcommerceCrawlStrategy(str, Enum):
     ignored. Use this when some subCategory links are misidentified by
     ML-extraction."""
 
+    none: str = "none"
+    """Treat input URLs as direct links to product detail pages, and
```
BurnzZ, line 36: Shall we also change the `url` input to accept newline-delimited URL values similar to the ongoing Article spider?

BurnzZ: Ahh sorry just saw the other PRs

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -32,6 +32,10 @@ class EcommerceCrawlStrategy(str, Enum):
     ignored. Use this when some subCategory links are misidentified by
     ML-extraction."""
 
+    none: str = "none"
```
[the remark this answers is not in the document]
Gallaecio: I have no strong opinion on whether it is cleaner to have 2 classes vs 1 extra crawl strategy.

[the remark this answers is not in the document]
Gallaecio: > I'm thinking about calling it something else which can give more meaning, like "Direct URLs to Product".

Only in the UI, or also in code? (seems rather long for code).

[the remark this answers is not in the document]
BurnzZ: > I have no strong opinion on whether it is cleaner to have 2 classes vs 1 extra crawl strategy 

To clarify on this, 2 classes would also mean 2 templates.

> Only in the UI, or also in code? (seems rather long for code).

Yes, in the code, it could just be `direct_product`.

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -55,6 +59,13 @@ class EcommerceSpiderParams(BaseSpiderParams):
                         "Use this when some subCategory links are misidentified by ML-extraction."
                     ),
                 },
+                EcommerceCrawlStrategy.direct_product: {
+                    "title": "Direct URLs to Product",
```
kmike, line 63: We don’t support multiple URLs yet. We can either change it to “URL”, or wait until multiple URL input is available.

I’d probably prefer to wait until multiple URL input is available, because it doesn’t look very useful with a single URL. 

But there might be use cases, so no strong opinion.

proway2 (approved):

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -38,6 +38,10 @@ class EcommerceCrawlStrategy(str, Enum):
     """Follow pagination and product detail pages. Subcategory links are
     ignored."""
 
+    direct_product: str = "direct_product"
```
BurnzZ, line 41: I'm thinking if we should have this in the Base strategies since it could also be applicable to the upcoming articles. Maybe call it `direct_item` or something?

kmike: Renaming to direct_item makes sense to me, because other strategies also use the same names across strategies (e.g. full vs full_product).

kmike: I'm not sure that it'd be possible to have it in some kind of Base params or Base stategy, as spiders are different, but anyways, this can be addressed separately if needed.

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -122,14 +133,19 @@ def _init_extract_from(self):
             )
 
     def get_start_request(self, url):
+        callback = (
+            self.parse_product
+            if self.args.crawl_strategy == EcommerceCrawlStrategy.direct_product
+            else self.parse_navigation
+        )
         meta = {
             "crawling_logs": {"page_type": "productNavigation"},
```
kmike, line 142: is this correct for direct_product strategy?

Issue #36, opened 2024-02-08
Allow multiple URLs as input to the spiders


VMRuiz:
### Description:
The current ecommerce spider in the repository accepts a single input URL. This means that crawling different categories in a website requires the creation of multiple spiders. However, this approach is impractical for several reasons:

1. Handling a large number of inputs becomes tedious and challenging to monitor.
2. Merging datasets requires additional effort.
3. Running numerous small spiders is inefficient in resource utilization.

### Proposed Solution:
Implementing this feature can be achieved through various methods

1. Allow multiple URLs in the url field using a separator like |.
2. Enable the use of an HTTPs URL to a CSV file with one input URL per line. This extensible to other protocols like S3, or SFTP.
3. Integrate with external services such as Google Spreadsheets or Airtable.

My prefered solution is number 2 as is usually the simpler and more versatile of all of them. It allows to update the URLs set on the file without having to re-configure the spider and there is a huge number of options to host such files without getting handcuffed into a specific provider.

Gallaecio:
1 is friendlier with a no-code approach.

If we are OK with only supporting HTTP/HTTPS start requests, we could automatically interpret input URLs with a different protocol as links to lists of input URLs, which would allow to support both 1 and 2 with a single spider parameter. Or we could follow an approach similar to that of `pip` for VCS, e.g. if the protocol is prefixed with `urls+` we treat it as a link to a list of input URLs, supporting all protocols, even HTTP or HTTPS, to point to a URL list.

VMRuiz:
What if we add a checkbox to indicate the URL is a input file instead ? It seems easier to use than having to prefix `urls+`.

Gallaecio:
What if we had a separate `input_urls` field, in addition to `urls`, to specify the URLs that point to a list of URLs? We get the flexibility of `urls+` without a custom schema, and having `urls` + some boolean option does not seem that different from having 2 URLs options, in the end is 2 fields in both cases.

VMRuiz:
The downside is that now you have to validate that people is not using `url` and `ìnput_urls` at the same time. 
Also, I think they could easily be confused, resulting in a bad user experience.

VMRuiz:
Alternatively, we could have a single field and decide what to do with it based on the headers responses from a HEAD request: 

- If response is a txt file, treat it as `input_urls`
- If responses is a html, treat it as current `url` 

The downside is that this method requires one additional requests, and there could network / bans issues that affects its outcome.

[...]

PR #41, opened 2024-02-15
Add a seed URL parameter

Gallaecio:
Resolves the second part of https://github.com/zytedata/zyte-spider-templates/issues/36, i.e. loading a list of URLs from the specified start URL.

### DIFF
--- a/tests/test_ecommerce.py
+++ b/tests/test_ecommerce.py
@@ -392,8 +392,7 @@ def test_metadata():
                     "description": (
                         "Enable this option if the specified initial URL "
                         "points to a list of URLs to crawl, with 1 URL per "
-                        "line. Note: When using this option, crawling outside "
-                        "the domain of the list URLs will be allowed."
+                        "line."
                     ),
                     "type": "boolean",
                 },
--- a/tests/test_ecommerce.py
+++ b/tests/test_ecommerce.py
@@ -1,11 +1,11 @@
 import logging
 import re
-from unittest.mock import MagicMock, call
+from unittest.mock import MagicMock, call, patch
 
 import pytest
+import requests
 import scrapy
 from pydantic import ValidationError
-from scrapy.http import TextResponse
 from scrapy_poet import DummyResponse
 from scrapy_spider_metadata import get_spider_metadata
 from zyte_common_items import ProbabilityRequest, Product, ProductNavigation, Request
@@ -652,25 +652,22 @@ def test_use_url_lists():
     spider = EcommerceSpider.from_crawler(crawler, url=url)
     start_requests = list(spider.start_requests())
     assert len(start_requests) == 1
-    assert start_requests[0].callback == spider.parse_navigation
+    assert start_requests[0].url == url
 
     spider = EcommerceSpider.from_crawler(crawler, url=url, use_url_lists=False)
     start_requests = list(spider.start_requests())
     assert len(start_requests) == 1
-    assert start_requests[0].callback == spider.parse_navigation
+    assert start_requests[0].url == url
 
-    spider = EcommerceSpider.from_crawler(crawler, url=url, use_url_lists=True)
+    with patch("zyte_spider_templates.spiders.ecommerce.requests.get") as mock_get:
+        response = requests.Response()
+        response._content = (
+            b"https://a.example\n \nhttps://b.example\nhttps://c.example\n\n"
+        )
+        mock_get.return_value = response
+        spider = EcommerceSpider.from_crawler(crawler, url=url, use_url_lists=True)
     start_requests = list(spider.start_requests())
-    assert len(start_requests) == 1
-    assert start_requests[0].callback == spider.parse_url_list
-
-    response = TextResponse(
-        url,
-        body=b"https://a.example\n \nhttps://b.example\nhttps://c.example\n\n",
-    )
-    requests = list(start_requests[0].callback(response))
-    assert len(requests) == 3
-    assert all(request.callback == spider.parse_navigation for request in requests)
-    assert requests[0].url == "https://a.example"
-    assert requests[1].url == "https://b.example"
-    assert requests[2].url == "https://c.example"
+    assert len(start_requests) == 3
+    assert start_requests[0].url == "https://a.example"
+    assert start_requests[1].url == "https://b.example"
+    assert start_requests[2].url == "https://c.example"
--- a/tests/test_ecommerce.py
+++ b/tests/test_ecommerce.py
@@ -459,23 +459,17 @@ def test_metadata():
                     "default": "",
                     "group": "inputs",
                 },
-                "seed_urls": {
-                    "anyOf": [
-                        {"type": "array", "items": {"type": "string"}},
-                        {"type": "null"},
-                    ],
-                    "title": "Seed URLs",
+                "seed_url": {
+                    "type": "string",
+                    "title": "Seed URL",
                     "description": (
-                        "URLs that point to lists with the initial URLs for "
-                        "the crawl. Both the list of seed URLs and the seed "
-                        "URL lists must be URLs separated by new lines. Enter "
-                        "the full URLs including http(s), you can copy and "
-                        "paste them from your browser. Example: "
-                        "https://toscrape.com/"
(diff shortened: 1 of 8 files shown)


zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -95,7 +106,10 @@ class EcommerceSpider(Args[EcommerceSpiderParams], BaseSpider):
     @classmethod
     def from_crawler(cls, crawler: Crawler, *args, **kwargs) -> scrapy.Spider:
         spider = super(EcommerceSpider, cls).from_crawler(crawler, *args, **kwargs)
-        spider.allowed_domains = [get_domain(spider.args.url)]
+        if not spider.args.use_url_lists:
+            spider.allowed_domains = [get_domain(spider.args.url)]
+        else:
+            pass  # https://github.com/scrapy/scrapy/issues/3257
```
Gallaecio, line 112: @kmike Shall I work upstream on this next?

Gallaecio: Worked around by loading URL lists synchronously.

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -102,7 +111,16 @@ class EcommerceSpider(Args[EcommerceSpiderParams], BaseSpider):
     @classmethod
     def from_crawler(cls, crawler: Crawler, *args, **kwargs) -> scrapy.Spider:
         spider = super(EcommerceSpider, cls).from_crawler(crawler, *args, **kwargs)
-        spider.allowed_domains = [get_domain(spider.args.url)]
+        if spider.args.use_url_lists:
+            response = requests.get(spider.args.url)
```
VMRuiz, line 115: Rather than using an external library, why don't we use Scrapy downloader handlers? 
That would also open the door for fetching s3 / FTP urls. I implemented something like that in the past:
https://github.com/VMRuiz/GenericCrawlerPOC/blob/main/generic_crawler/input.py

VMRuiz: On my example, calling the handlers skips the middlewares, so offsite requests should not be an issue. 
Although I admit that the resulting code is a little bit complex and collides with Zyte API usage if enabled.

Gallaecio: Just to clarify, the issue with allowed domains is that we set it automatically based on start URLs from `from_crawler`. If we need to post-pone the fetching of start URLs to the asynchronous `start_requests`, by the time that method is called the offsite middleware will have already processed the list of allowed domains.

We could hack around it (find the middleware instance and edit it from `start_requests`), it is just rather ugly. Or we can *not* set a limit on allowed domains to begin with when loading lists of URLs.

proway2 (approved):

Gallaecio:
[Switched to a single seed URL](https://github.com/zytedata/zyte-spider-templates/pull/41/commits/e31b784fef2ee811d6ce1a1979759b7f617e95e1) after feedback from @VMRuiz.

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -92,7 +93,16 @@ class EcommerceSpider(Args[EcommerceSpiderParams], BaseSpider):
     @classmethod
     def from_crawler(cls, crawler: Crawler, *args, **kwargs) -> scrapy.Spider:
         spider = super(EcommerceSpider, cls).from_crawler(crawler, *args, **kwargs)
-        spider.allowed_domains = [get_domain(spider.args.url)]
+        seed_url = spider.args.seed_url
+        if seed_url:
+            response = requests.get(seed_url)
+            urls = [url.strip() for url in response.text.split("\n")]
+            urls = [url for url in urls if url]
```
kmike, line 100: Would you mind moving parsing logic to a function, for easier testing & documentation?

kmike (approved):

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -92,7 +93,16 @@ class EcommerceSpider(Args[EcommerceSpiderParams], BaseSpider):
     @classmethod
     def from_crawler(cls, crawler: Crawler, *args, **kwargs) -> scrapy.Spider:
         spider = super(EcommerceSpider, cls).from_crawler(crawler, *args, **kwargs)
-        spider.allowed_domains = [get_domain(spider.args.url)]
+        seed_url = spider.args.seed_url
+        if seed_url:
+            response = requests.get(seed_url)
```
BurnzZ, line 98: Correct me if I'm wrong but would the `crawler` instance here already have access to a good-to-go Downloader? If so, can we use it instead of using the external `requests` library?

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -92,7 +93,16 @@ class EcommerceSpider(Args[EcommerceSpiderParams], BaseSpider):
...
+            urls = [url.strip() for url in response.text.split("\n")]
```
BurnzZ, line 99: Perhaps we should also validate URLs here using the _URL_PATTERN and maybe drop any malformed entries in the process.

Though maybe a better approach would be to error out and don't start the crawl since depending on the use case, it may be a waste to begin a crawl if even a single malformed URL was present (_due to human error)_.

zyte_spider_templates/spiders/ecommerce.py
Gallaecio: The goal is for the request to be handled synchronously. Doing otherwise would complicate the code a bit.

tests/test_ecommerce.py
```diff
@@ -371,16 +372,42 @@ def test_metadata():
...
                         "Initial URL for the crawl. Enter the full URL including http(s), "
                         "you can copy and paste it from your browser. Example: https://toscrape.com/"
                     ),
+                    "exclusiveRequired": True,
+                    "group": "inputs",
                     "pattern": r"^https?://[^:/\s]+(:\d{1,5})?(/[^\s]*)*(#[^\s]*)?$",
                     "title": "URL",
                     "type": "string",
                 },
+                "seed_url": {
```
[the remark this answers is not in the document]
BurnzZ: > Specially since whether or not the URLs are seed URLs would determine whether or not we try automatic extraction on them.

Ahhh this made me realize that the PO would have `response: AnyResponse` _(i.e. for the first request's response)_, and if `response` is a `BrowserResponse`, we don't have the response headers to check. Thus, it'd be tricky to do the implicit approach via `"Content-Type": "text/plain"` header checks. 🙅‍♂️

> Or implement some client-side syntax to indicate that a URL is a seed URL,

I would avoid some non-standard syntax that needs to be learned by the user here.

-------

One last thing to discuss and sort out: _What about the use case of a user trying to crawl multiple seed URLs?_

For example, the user has two or more pre-defined textfiles which are a bunch of websites per country:

- https://somesource.com/fr-websites.txt
  - https://fr.example-1.com
  - https://fr.example-2.com
  - https://fr.example-3.com
  - https://fr.example-4.com
- https://somesource.com/de-websites.txt
  - https://de.example-1.com
  - https://de.example-2.com
  - https://de.example-3.com
- https://somesource.com/es-websites.txt
  - ...
- ...

And say, they want to make a virtual spider that crawls EMEA websites.

In order to support it, they can re-use the existing textfiles they have to crawl them by passing them to the `urls` param:

- https://somesource.com/fr-websites.txt
- https://somesource.com/de-websites.txt
- https://somesource.com/es-websites.txt
- ...

Moreover, there's a `boolean` flag that they need to turn on to treat such URLs as seeds.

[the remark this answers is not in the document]
Gallaecio: I suggested the boolean approach originally, but @VMRuiz believes that, when using seed URLs, the bast majority of cases will use a single seed URL. Which I did not realize at the beginning, but makes sense.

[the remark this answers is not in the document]
BurnzZ: Yeah, I don't think this would be a huge dealbreaker to users as well, as they can simply make a new list with the combined URLs. 

Glad all other use cases were explored!

BurnzZ (approved):

PR #46, opened 2024-03-20
Define standard parameters through individual mixins

Gallaecio:
Alternative to https://github.com/zytedata/zyte-spider-templates/pull/34 using mixins.

### DIFF
--- a/docs/reference/index.rst
+++ b/docs/reference/index.rst
@@ -2,33 +2,40 @@
 Reference
 =========
 
-Base classes
-============
+Spiders
+=======
 
-.. autopydantic_model:: zyte_spider_templates.spiders.base.BaseSpiderParams
-    :inherited-members: BaseModel
+.. autoclass:: zyte_spider_templates.BaseSpider
 
-.. autoclass:: zyte_spider_templates.spiders.base.BaseSpider
+.. autoclass:: zyte_spider_templates.EcommerceSpider
 
-.. autoenum:: zyte_spider_templates.spiders.base.ExtractFrom
-    :noindex:
-
-.. autoenum:: zyte_spider_templates.spiders.base.Geolocation
-    :noindex:
-
-E-commerce
-==========
-
-.. autopydantic_model:: zyte_spider_templates.spiders.ecommerce.EcommerceSpiderParams
-    :noindex:
-    :inherited-members: BaseModel
-
-.. autoenum:: zyte_spider_templates.spiders.ecommerce.EcommerceCrawlStrategy
-    :noindex:
-
-.. autoclass:: zyte_spider_templates.spiders.ecommerce.EcommerceSpider
 
 Pages
 =====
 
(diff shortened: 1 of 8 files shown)


zyte_spider_templates/spiders/base.py
```diff
@@ -1,85 +1,24 @@
...
-)
-from zyte_spider_templates.documentation import document_enum
+from ..params import ExtractFromParam, GeolocationParam, MaxRequestsParam, UrlParam
 
 # Higher priority than command-line-defined settings (40).
 ARG_SETTING_PRIORITY: int = 50
 
 
-@document_enum
-class ExtractFrom(str, Enum):
```
Gallaecio, line 20: Both `Geolocation` and `ExtractFrom` would previously be imported from here and are now imported from `zyte_spider_templates.params`.

I was hoping to deprecate them properly, but I had to go with a backward-incompatible change, because I could not think of a proper way of deprecating them that was not [rather hacky](https://stackoverflow.com/a/7668273/939364), given (TIL) `Enum` subclassing is not possible.

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -67,6 +68,10 @@ class EcommerceSpiderParams(BaseSpiderParams):
     )
 
 
+class EcommerceSpiderParams(EcommerceCrawlStrategyParam, BaseSpiderParams):
+    pass
```
kmike, line 72: Is the order of params the same as before, i.e. if you add a mixin in front, the param is added to the end?

zyte_spider_templates/spiders/base.py
```diff
@@ -1,85 +1,24 @@
...
-            },
-        },
-    )
+class BaseSpiderParams(
+    UrlParam,
+    GeolocationParam,
+    MaxRequestsParam,
+    ExtractFromParam,
+    BaseModel,
+):
```
kmike, line 20: Aesthetically, this looks fine - not as bad as I expected :)

zyte_spider_templates/spiders/ecommerce.py
kmike: In a previous example the order was the opposite - URL field first, etc. How does it work?

Gallaecio: https://github.com/zytedata/zyte-spider-templates/pull/46/commits/af9fe4d71bbf39c0e15f5d9001f82e10431c57f2

docs/reference/index.rst
```diff
@@ -2,33 +2,40 @@
 Reference
```
kmike, line 2: What do you think about adding usage docs for it as well in the same PR? 
https://zyte-spider-templates.readthedocs.io/en/latest/customization/spiders.html#customizing-parameters

kmike (approved):

PR #47, opened 2024-03-27
Article AI Spider

felipdc:
### Adds support for article webpages leveraging ZyteAPI Extract for Article and Article Navigation.
- `Heuristics.py` updated for better reusability.
- Added docs for Article template spider.
- Added tests for Article template spider.


kmike:
Thanks @felipdc! I'm closing it, because we released an article template. 

I do like some of the changes here, e.g. the refactoring of the constants in heuristics.py.

PR #52, opened 2024-04-26
Combining `full` and `navigation` strategies

BurnzZ:
# Overview

This combines the current `full` and `navigation` crawl strategies into a single one, wherein:
- if the given URL is detected to be a homepage, then we use a `full` crawl for it. However, we only discover new links via heuristics on the homepage and not on succeeding "_productNavigation-heuristic_" pages.
- if the given URL doesn't seem to b a homepage, then we perform the usual `navigation` crawl.

Some benefits for this:
- Since this replaces the default crawl strategy which is currently set to `navigation`, users that provide the homepage as the starting URL can benefit from it if `subCategory` links are missed out during Zyte API extraction.
- This also help prevent human errors where the incorrect crawl strategy was inadvertently used since it should be smart enough to decide which is the best one.
- Some experiments show that there's some small amount of increased crawling quality. That is, because it now avoids crawling other links discovered by heuristics when outside of the homepage, there's less Product Navigation requests performed which leads to less False Positive extracted products _(which are then later dropped due to low probability values)_.

----

TODO:
- [x] name for the combined strategy. ~Called it `default` for now.~ Renamed to `automatic`.
- [x] decision: what's our backwards-compatibility approach for deprecating the other strategies?
- [x] a more robust way to detect homepages
- [x] confirm UI handling of default changes (i.e. create a virtual spider with default crawl strategy, deploy new code, verify the old virtual spider uses the old default when run)


zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -23,15 +24,11 @@
 
 @document_enum
 class EcommerceCrawlStrategy(str, Enum):
-    full: str = "full"
-    """Follow most links within the domain of URL in an attempt to discover and
-    extract as many products as possible."""
-
-    navigation: str = "navigation"
+    default: str = "default"
```
kmike, line 27: I think we should start with keeping it backwards compatible - offer all 3 strategies for now.

kmike: We can handle deprecation later.

kmike: Or does it require too much noise in the spider?

BurnzZ: Updated the PR to still include the existing `full` + `navigation` strategies but then updated their description to mention they're deprecated.

tests/test_ecommerce.py
```diff
@@ -465,18 +465,30 @@ def test_metadata():
...
+                                "Follow pagination, subcategories, and product detail pages. "
+                                "If starting on a homepage, it would attempt to discover other "
+                                "URLs in the page using heuristics."
+                            ),
+                            "title": "Default",
+                        },
                         "full": {
-                            "description": "Follow most links within the domain of URL in an attempt to discover and extract as many products as possible.",
+                            "description": (
+                                "(Deprecated. Use Default instead) Follow most links "
```
kmike, line 481: I'm not even sure we should be deprecating those 2 extra strategies :) What do you think about not getting them deprecated now, and getting a bit more feedback?

BurnzZ: Ahhh this is what @Gallaecio has recently raised and discussed in the call today. 👍 Yeah, we could remove marking them as deprecated and still expose them to users. 

Adrian's main concern was about making `navigation` still available to users in case a given website works quite well with it _(without heuristics)_. If so, they could reduce the number of requests due to having less requests due to heuristics.

Gallaecio: @BurnzZ suggested an alternative approach, which I think is worth considering: have a separate boolean/enum field for heuristics.

So, we could make the crawl strategy be about the crawl strategy only, not heuristics. We could deprecate `full` (because in terms of crawl strategy, `full` is equivalent to `navigation` pointed to a homepage), and keep `navigation`, `pagination-only`, and a future `direct-item`. Then have a separate Heuristics field with options like `follow-all-homepage` (this new approach), `follow-all-navigation` (equivalent to the `full` old crawl strategy, should be default for backward compatibility at least in pre-existing virtual spiders), `none` (equivalent to the old `navigation`), `sitemap`, etc.

I think it would provide a better user experience overall, but it might make backward-compatibility a bit harder to accomplish.

kmike: Do we need to handle it in this PR? On a first sight, we can just add a new strategy for now, make it default, and then we can think about all the deprecations, backwards compatibility, new options, etc.

kmike: Would this approach make the transition any worse?

Gallaecio: I think and extra field would indeed negatively affect the transition. With the strategy field only, existing virtual spiders in Scrapy Cloud can maintain backward compatibility as long as Scrapy Cloud stored the actual value selected (even if it was the default), which seems to be the case.

With a new field that Scrapy Cloud knows nothing about, whose default value would cause a backward-incompatible behavior, keeping the old behavior in pre-existing virtual spiders in Scrapy Cloud would be much harder, I think.

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -23,6 +24,13 @@
 
 @document_enum
 class EcommerceCrawlStrategy(str, Enum):
+    default: str = "default"
```
kmike, line 27: I think we have 2 options for the names.

1. Name it "default". It means that if we would change the default later, we'll be changing behavior of this strategy. So, all the spiders will get the new default. 
2. Name it something like "combined", "full+navigation", I don't know. This way, it can still be default, and the name is not going to be visible anywhere, but if we decide to change the default behavior, then the switch won't be automatic anymore.

I'm not sure what's better :) Leaning towards the first approach, but the second may also be worth considering. @BurnzZ @Gallaecio any thought?

By the way, if we go with "default" approach, I also wonder if we should name it something like "automatic" instead. It feels a bit off to have a strategy named "default" when there is an option to not have it as a default (e.g. in a template subclass); i think it's better to use a name which describes what the strategy does.

BurnzZ: I like the new name "automatic" as well. 👍 

I'm not quite sure about naming it like "full+navigation" since this can change in the future. For example, heuristics might be smarter to avoid crawling the subCategory links if it detects they are mostly FPs. Thus, the strategy now also includes "pagination_only"

BurnzZ: Renamed to "automatic" https://github.com/zytedata/zyte-spider-templates/pull/52/commits/dc947d1140bf4058ca1bf67dd60572364b0a7b64

tests/test_ecommerce.py
```diff
@@ -465,19 +465,29 @@ def test_metadata():
...
                 "crawl_strategy": {
-                    "default": "full",
+                    "default": "automatic",
                     "description": "Determines how the start URL and follow-up URLs are crawled.",
                     "enumMeta": {
+                        "automatic": {
+                            "description": (
+                                "Follow pagination, subcategories, and product detail pages. "
+                                "If starting on a homepage, it would attempt to discover other "
+                                "URLs in the page using heuristics."
```
kmike, line 475: Do you think we'll extend the heuristics in the future, while keeping the name?

If so, it could make sense to document this strategy as "automatic". We may still explain the exact behavior, as an implementation detail which can change.

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -266,3 +288,17 @@ def get_parse_product_request(
         )
         scrapy_request.meta["allow_offsite"] = True
         return scrapy_request
+
+    def page_params_for_heuristics(
+        self, response: DummyResponse
```
kmike, line 293: what do you think about passing page_params here? Or maybe response.meta. It looks weird that we require DummyResponse; in most cases parameters annotated as DummyResponse shouldn't be used in the body.

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -266,3 +288,17 @@ def get_parse_product_request(
         )
         scrapy_request.meta["allow_offsite"] = True
         return scrapy_request
+
+    def page_params_for_heuristics(
```
kmike, line 292: Would it make sense to make this method private?

tests/test_ecommerce.py
BurnzZ: Good call. Yeah, we'll definitely have some changes soon like some good utilization of sitemaps for better item coverage.

What do you think of this new one? https://github.com/zytedata/zyte-spider-templates/pull/52/commits/ad2776ac41f2bd8f0cf871ff6450f1e5e0c8cfe2

zyte_spider_templates/spiders/ecommerce.py
BurnzZ: That's much better! https://github.com/zytedata/zyte-spider-templates/pull/52/commits/eb98f4e0bfbe02c2dfa72b001b4d400378a6c914

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -125,7 +146,10 @@ def get_start_request(self, url):
         meta = {
             "crawling_logs": {"page_type": "productNavigation"},
         }
-        if self.args.crawl_strategy == EcommerceCrawlStrategy.full:
+        if self.args.crawl_strategy == EcommerceCrawlStrategy.full or (
+            self.args.crawl_strategy == EcommerceCrawlStrategy.automatic
+            and is_homepage(url)
```
kmike, line 151: What do you think about adding a log message, so that from the logs it's clear if a spider thinks a seed URL is a home page or not?

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -266,3 +290,16 @@ def get_parse_product_request(
         )
         scrapy_request.meta["allow_offsite"] = True
         return scrapy_request
+
+    def _page_params_for_heuristics(
```
kmike, line 294: by the name, I wouldn't expect the page_params to be modified. What do you think about either making a copy, or renaming a method (something like `_modify_page_params_for_heuristics`)?

BurnzZ: I was initially thinking about making a copy but given that the `page_params` could be arbitrarily nested, using `copy.deepcopy` might take a hit in the performance. 

Renaming it to `_modify_page_params_for_heuristics` seems safer. https://github.com/zytedata/zyte-spider-templates/pull/52/commits/19604aa1476a9876f1b65a473b5ac2aa219ae172

zyte_spider_templates/spiders/ecommerce.py
BurnzZ: Sounds good! https://github.com/zytedata/zyte-spider-templates/pull/52/commits/bffc9593daa5700294e99072c330b02225c9f7b8

kmike: * I think it's best to have logging in the spider, not in the utility function. 
* Not sure about this: should it be DEBUG or INFO? How are similar messages logged?

zyte_spider_templates/heuristics.py
```diff
@@ -56,3 +65,44 @@ def might_be_category(url: str) -> bool:
...
+    "",
+    "/index",
+    "/index.html",
+    "/index.htm",
+    "/index.php",
+    "/home",
+}
+
+
+def is_homepage(url: str) -> bool:
```
kmike, line 80: hey @BurnzZ!

I've run an experiment:

1. Got a list of popular and less popular domains (~160K)
2. Created a spider, which just downloads the front page, following all redirects.
3. Then, if after all redirects the URL is not barebone `/`, stored created an item.

So, this way we can get a dataset of positive homepages. It's a bit simplistic - there could be homepage variants with different URLs, and it won't help with catching false positives (that's a separate question, how to get a dataset for this).

In the end, I've got ~10K example non-trivial homepage URLs: https://gist.github.com/kmike/bcc970549e428ee5518868988ebd39dd

Could you please check the heuristics against this list? 

As currently we don't have a dataset to catch false positives, we need to be careful and conservative with adding more heuristics, but it seems there could be some common patterns which can be added to improve the results

zyte_spider_templates/spiders/ecommerce.py
BurnzZ: What do you think about this? https://github.com/zytedata/zyte-spider-templates/pull/52/commits/b4ffc303e6380e307ab8fb253e8463401ff8a9ef

I think adding a bit of explanation to the automatic strategy would help in the long run, especially when we're considering other ways to add more heuristics like sitemaps later on.

We can use INFO so the logs would be available by default to Scrapy Cloud users. It's also similar to how the other logs in the spider where it uses INFO to indicate/clarify some spider behaviors.

zyte_spider_templates/heuristics.py
BurnzZ: So far, the `is_homepage()` heuristic detects around 50% of the URLs as homepages. It's usually the case of having these suffix patterns in the URL:

- `/de/`
- `/en`
- `/home`
- `/en-US`
- `/en-us`
- `/index.php`
- `/us/en`
- `/pt_br/index`
- `/it_IT`
- `/en/home/`
- `/en/index.html?`

The other 50% of the URLs which were deemed to not be URLs go something like:

- `/content/boaml/en_us/home.html`
- `/DIGITAL/Navigation/DE/Home/home.html`
- `/all-about/oxfordshire`
- `/en/services/health.html`
- `/forums/index.php`
- `?gws_rd=ssl` - or basically any URL with Query params 
- `/en/games`
- `/portal`
- `/archives/main/`
- `/pages/homepage/`
- `/Pages/Default.aspx`
- `/en/page/home`

So far, I think the current setup of the `is_homepage()` heuristics is conservative enough. Most of them were False Negatives as can be observed from 50% of the results not detected as homepages.

zyte_spider_templates/heuristics.py
```diff
@@ -56,3 +63,43 @@ def might_be_category(url: str) -> bool:
...
+    url_path = url_split.path.rstrip("/").lower()
+
+    # Finds and removes URL subpaths like "/us/en", "en-us", "en-uk", etc.
+    if _url_has_locale_pair(url_path):
+        url_path = url_path[6:]
+
+    # Finds and removes URL subpaths like "/en", "/fr", etc.
+    match = re.search(r"/(\w{2})(?!\w)", url_path)
+    if match and (match.group(1) in LANG_CODES or match.group(1) in COUNTRY_CODES):
+        url_path = url_path[3:]
```
Gallaecio, line 90: I’m curious if this is more efficient than building a regular expression out of `LANG_CODES` and `COUNTRY_CODES` instead.

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -266,3 +302,16 @@ def get_parse_product_request(
...
+    def _modify_page_params_for_heuristics(
+        self, page_params: Optional[Dict]
+    ) -> Dict[str, Any]:
+        page_params = page_params or {}
+        # Only allow heuristic extraction of links in non-homepage when on "full" crawl.
+        if (
+            self.args.crawl_strategy != EcommerceCrawlStrategy.full
+            and "full_domain" in page_params
+        ):
+            page_params.pop("full_domain")
```
Gallaecio, line 315: :lipstick: 

```suggestion
        if self.args.crawl_strategy != EcommerceCrawlStrategy.full:
            page_params.pop("full_domain", None)
```

zyte_spider_templates/heuristics.py
BurnzZ: Good thought there. I've benchmarked this and found out that they're the same.

zyte_spider_templates/heuristics.py
```diff
@@ -56,3 +63,43 @@ def might_be_category(url: str) -> bool:
                 return False
 
     return True
+
+
+INDEX_URL_PATHS = {
```
wRAR, line 68: Do we plan to improve this list later?

wRAR (approved):

zyte_spider_templates/heuristics.py
BurnzZ: Yeah, we can update it as we encounter new ones.

PR #53, opened 2024-04-30
Release notes for 0.7.2

Gallaecio:
@VMRuiz requested a release with parameter mixins. I branched out 281f1f5 into [0.7.x](https://github.com/zytedata/zyte-spider-templates/commits/0.7.x) to exclude more recent changes that we don’t want to release yet.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,18 @@
 Changes
 =======
 
+0.7.2 (2024-MM-YY)
+------------------
+
+* Implemented :ref:`mixin classes for spider parameters <parameter-mixins>`, to
+  improve reuse.
+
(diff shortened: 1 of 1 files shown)


kmike:
What prevents us from releasing main branch?

Gallaecio:
Yes, the seed URL field would just appear below the URL one, it would not be visually clear that you cannot combine both. We could release anyway, it’s not a big issue, but it’s not pretty.

kmike:
> Yes, the seed URL field would just appear below the URL one, it would not be visually clear that you cannot combine both. 

Thanks! I think the way you did it here is the best, +1 to proceed with a branch.

kmike (approved):

PR #55, opened 2024-07-19
Add LocationParam

VMRuiz:


VMRuiz:
There seems to be some issues with mypy typing. Could you please recommend what's the best way to fix it?

VMRuiz:
I added some exceptions in tests code for mypy errors because I couldn't make them work in the class definition itself. 
Also, there is another error about a licence in twine pipeline

wRAR:
The twine problem is global and should be fixed by bumping the deps in `twinecheck`

Gallaecio (approved):
I guess docs will make more sense once we actually implement a location parameter for spiders.

PR #62, opened 2024-08-27
SERP (MVP)

Gallaecio:
Pagination is currently hard-coded to 10, which is not always accurate I believe. However, the next iteration of SERP extraction will include a nextPage entry, so I think it is best to start with this simple approach and switch to a nextPage-based implementation later.

This implementation does not rely on scrapy-poet because scrapy-zyte-api does not yet support `serp` properly. To be changed once we add proper support for it there (working on it).

### DIFF
--- a/zyte_spider_templates/spiders/serp.py
+++ b/zyte_spider_templates/spiders/serp.py
@@ -1,0 +1,135 @@
+from typing import Any, Dict, Iterable
+
+import requests
+import scrapy
+from pydantic import BaseModel, ConfigDict, Field, model_validator
+from scrapy import Request
+from scrapy.crawler import Crawler
+from scrapy_spider_metadata import Args
+from w3lib.url import add_or_replace_parameter
+from zyte_common_items import Serp
+
+from zyte_spider_templates.spiders.base import BaseSpider
+from zyte_spider_templates.utils import get_domain
+
+from ..params import MaxRequestsParam, UrlParam, UrlsFileParam, UrlsParam
+from ..utils import load_url_list
+from .base import _INPUT_FIELDS
+
+
+class SerpMaxPagesParam(BaseModel):
+    max_pages: int = Field(
+        title="Pages",
+        description="Maximum number of result pages to visit per input URL.",
+        default=1,
+    )
+
+
+class SerpSpiderParams(
+    MaxRequestsParam,
+    SerpMaxPagesParam,
+    UrlsFileParam,
+    UrlsParam,
+    UrlParam,
+    BaseModel,
+):
+    model_config = ConfigDict(
+        json_schema_extra={
+            "groups": [
+                {
+                    "id": "inputs",
+                    "title": "Inputs",
+                    "description": (
+                        "Input data that determines the start URLs of the crawl."
+                    ),
+                    "widget": "exclusive",
+                },
+            ],
+        },
+    )
+
+    @model_validator(mode="after")
+    def single_input(self):
+        """Fields
+        :class:`~zyte_spider_templates.spiders.ecommerce.EcommerceSpiderParams.url`
+        and
+        :class:`~zyte_spider_templates.spiders.ecommerce.EcommerceSpiderParams.urls_file`
+        form a mandatory, mutually-exclusive field group: one of them must be
+        defined, the rest must not be defined."""
+        input_fields = set(
+            field for field in _INPUT_FIELDS if getattr(self, field, None)
+        )
+        if not input_fields:
+            input_field_list = ", ".join(_INPUT_FIELDS)
+            raise ValueError(
+                f"No input parameter defined. Please, define one of: "
+                f"{input_field_list}."
+            )
+        elif len(input_fields) > 1:
+            input_field_list = ", ".join(
+                f"{field} ({getattr(self, field)!r})" for field in input_fields
+            )
+            raise ValueError(
+                f"Expected a single input parameter, got {len(input_fields)}: "
+                f"{input_field_list}."
+            )
+        return self
+
+
+class SerpSpider(Args[SerpSpiderParams], BaseSpider):
+    """Yield results from search engine result pages (SERP).
+
+    See :class:`~zyte_spider_templates.spiders.ecommerce.SerpSpiderParams`
+    for supported parameters.
+
+    .. seealso:: :ref:`serp`.
+    """
+
+    name = "serp"
+
+    metadata: Dict[str, Any] = {
+        **BaseSpider.metadata,
+        "title": "SERP",
+        "description": "Template for spiders that extract search engine results.",
+    }
+
+    @classmethod
+    def from_crawler(cls, crawler: Crawler, *args, **kwargs) -> scrapy.Spider:
+        spider = super().from_crawler(crawler, *args, **kwargs)
+        spider._init_input()
+        return spider
+
+    def _init_input(self):
+        urls_file = self.args.urls_file
+        if urls_file:
+            response = requests.get(urls_file)
+            urls = load_url_list(response.text)
+            self.logger.info(f"Loaded {len(urls)} initial URLs from {urls_file}.")
+            self.start_urls = urls
+        elif self.args.urls:
+            self.start_urls = self.args.urls
+        else:
+            self.start_urls = [self.args.url]
+        self.allowed_domains = list(set(get_domain(url) for url in self.start_urls))
+
+    def get_start_request(self, url):
+        return Request(
+            url=url,
+            callback=self.parse_serp,
+            meta={
+                "crawling_logs": {"page_type": "serp"},
+                "zyte_api": {
+                    "serp": True,
+                },
+            },
+        )
+
+    def start_requests(self) -> Iterable[Request]:
+        for url in self.start_urls:
+            for start in range(0, self.args.max_pages * 10, 10):
+                if start:
+                    url = add_or_replace_parameter(url, "start", str(start))
+                yield self.get_start_request(url)
+
+    def parse_serp(self, response) -> Iterable[Serp]:
+        yield Serp.from_dict(response.raw_api_response["serp"])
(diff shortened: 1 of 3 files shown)


zyte_spider_templates/spiders/serp.py
```diff
@@ -0,0 +1,135 @@
...
+            ],
+        },
+    )
+
+    @model_validator(mode="after")
+    def single_input(self):
+        """Fields
+        :class:`~zyte_spider_templates.spiders.ecommerce.EcommerceSpiderParams.url`
+        and
+        :class:`~zyte_spider_templates.spiders.ecommerce.EcommerceSpiderParams.urls_file`
```
wRAR, line 56: Class name needs to be updated I think

zyte_spider_templates/spiders/serp.py
```diff
@@ -0,0 +1,135 @@
...
+                f"Expected a single input parameter, got {len(input_fields)}: "
+                f"{input_field_list}."
+            )
+        return self
+
+
+class SerpSpider(Args[SerpSpiderParams], BaseSpider):
+    """Yield results from search engine result pages (SERP).
+
+    See :class:`~zyte_spider_templates.spiders.ecommerce.SerpSpiderParams`
```
wRAR, line 82: Old module name

wRAR (approved):

zyte_spider_templates/spiders/serp.py
```diff
@@ -0,0 +1,131 @@
...
+from ._google_domains import GoogleDomain
+from .base import BaseSpider
+
+
+class SearchKeywordsParam(BaseModel):
+    search_keywords: Optional[List[str]] = Field(
+        title="Search Keywords",
+        description=(
+            "Keywords to search for. Use multiple lines to trigger multiple "
+            "searches for different search keywords."
```
kmike, line 20: It feels a bit confusing - if a single line in this item is "search keywords", then why is the whole field called "search keywords"? Would it make sense to rename the field, e.g. to "search_queries"? In this case, it's going to be a search query per line. 

"Keywords" also could be a bit imprecise. I think it implies that the order of search keywords in a query doesn't matter much, and they're not necessarily related to each other - it's just a bunch of keywords to look up in the index. I'm not sure queries are treated like this by search engines now.

Gallaecio: https://github.com/zytedata/zyte-spider-templates/pull/62/commits/c3a2f2369ce34aa1041fffd60a066396b77b59f9 ?

zyte_spider_templates/spiders/serp.py
```diff
@@ -0,0 +1,131 @@
...
+from ._google_domains import GoogleDomain
+from .base import BaseSpider
+
+
+class SearchQueriesParam(BaseModel):
+    search_queries: Optional[List[str]] = Field(
+        title="Search Queries",
+        description=(
+            "Input 1 search query per line. A search query is a string of "
+            "search keywords (e.g. foo bar)."
```
kmike, line 20: @Gallaecio as we discussed on a call, probably better to avoid mentioning keywords

CHANGES.rst
```diff
@@ -1,6 +1,12 @@
 Changes
 =======
 
+Unreleased
+----------
+
+* Now requires ``zyte-common-items >= 0.22.0``.
```
kmike, line 7: should it mention SERP spider?

kmike (approved):

kmike:
With the serp item released, what are the obstacles in adding Serp support to zyte-common-items?

CHANGES.rst
Gallaecio: Yes, I just added that bit because I was afraid we could miss it when writing the notes. I don’t think we will miss the spider part :slightly_smiling_face:

Gallaecio:
I’ve just opened https://github.com/scrapy-plugins/scrapy-zyte-api/pull/218 out of local changes I had started. There is no blocker, but it is not entirely a trivial change, it is not very complex either, but the special behavior of serp compared to other data types does complicate things more than usual, and there are other Zyte API features that are not covered in scrapy-zyte-api yet that we should also cover while adding serp support.

zyte_spider_templates/spiders/serp.py
```diff
@@ -0,0 +1,128 @@
...
+                "zyte_api.aggressive_retrying",
+                priority="spider",
+            )
+
+    def get_start_request(self, url):
+        return Request(
+            url=url,
+            callback=self.parse_serp,
+            meta={
+                "crawling_logs": {"page_type": "serp"},
```
kmike, line 105: I seems the resulting crawl logs don't quite make sense for serp:

```
Number of Requests per page type:
- product: 0
- nextPage: 0
- subCategories: 0
- productNavigation: 0
- productNavigation-heuristics: 0
- unknown: 0
Structured Logs:
{
  "time": "2024-09-13 12:47:31",
  "current": {
    "url": "https://www.google.com/search?q=frappe",
    "request_url": "https://www.google.com/search?q=frappe",
    "request_fingerprint": "231822665b55ba64d4f531b734de660fe1b81b5d",
    "page_type": "serp",
    "probability": null
  },
  "to_crawl": {
    "product": [],
    "nextPage": [],
    "subCategories": [],
    "productNavigation": [],
    "productNavigation-heuristics": [],
    "unknown": []
  }
}
```

Gallaecio:
Now:

```
Nothing to crawl.
Structured Logs:
{
  "time": "2024-09-17 00:12:01",
  "current": {
    "url": "https://www.google.com/search?q=foo&start=10",
    "request_url": "https://www.google.com/search?q=foo&start=10",
    "request_fingerprint": "1b95b08d33b7b8401ec2c6126fc1399100ab12b9",
    "page_type": "serp",
    "probability": null
  },
  "to_crawl": {}
}
```

zyte_spider_templates/spiders/serp.py
Gallaecio: Now:

```
Nothing to crawl.
Structured Logs:
{
  "time": "2024-09-17 00:12:01",
  "current": {
    "url": "https://www.google.com/search?q=foo&start=10",
    "request_url": "https://www.google.com/search?q=foo&start=10",
    "request_fingerprint": "1b95b08d33b7b8401ec2c6126fc1399100ab12b9",
    "page_type": "serp",
    "probability": null
  },
  "to_crawl": {}
}
```

Gallaecio:
@BurnzZ Any quick thought on the crawl logging changes?

zyte_spider_templates/spiders/serp.py
```diff
@@ -0,0 +1,128 @@
...
+
+    @field_validator("search_queries", mode="before")
+    @classmethod
+    def validate_search_queries(cls, value: Union[List[str], str]) -> List[str]:
+        """Validate a list of search queries.
+        If a string is received as input, it is split into multiple strings
+        on new lines.
+        """
+        if isinstance(value, str):
+            value = value.split("\n")
```
BurnzZ, line 32: I think we should put this at the bottom after the `if not value` check since having a input value of `""` would result in `[""]` which fails the non-empty check below. We'd want the `ValueError` that says input is empty.

zyte_spider_templates/spiders/serp.py
```diff
@@ -0,0 +1,128 @@
...
+        if not value:
+            raise ValueError("The search_queries parameter value is missing or empty.")
+        result = []
+        for v in value:
+            if not (v := v.strip()):
+                continue
+            result.append(v)
```
BurnzZ, line 39:
```suggestion
            if v := v.strip():
                result.append(v)
```
minor

zyte_spider_templates/spiders/serp.py
```diff
@@ -0,0 +1,128 @@
...
+        return result
+
+
+class SerpMaxPagesParam(BaseModel):
+    max_pages: int = Field(
+        title="Pages",
+        description="Maximum number of result pages to visit per input URL.",
```
BurnzZ, line 46:
```suggestion
        description="Maximum number of result pages to visit per search query input.",
```

zyte_spider_templates/spiders/serp.py
```diff
@@ -0,0 +1,128 @@
...
+        for v in value:
+            if not (v := v.strip()):
+                continue
+            result.append(v)
+        return result
+
+
+class SerpMaxPagesParam(BaseModel):
+    max_pages: int = Field(
+        title="Pages",
```
BurnzZ, line 45:
```suggestion
        title="Max Pages",
```

BurnzZ (approved):
Hi @Gallaecio , I just tested and reviewed the changes pertaining to the crawling logs. All looks well. 👍 

Left a few minor comments as well. Aside from that, LGTM!

zyte_spider_templates/spiders/serp.py
Gallaecio: Great catch. But in order to also cause `\n` to become `[]`, I will instead move the exception later, to check the contents of `result` instead.

zyte_spider_templates/spiders/serp.py
Gallaecio: What if we take “input” away as well?

Gallaecio:
Release notes added, I’ll handle the release as soon as we merge.

zyte_spider_templates/spiders/serp.py
BurnzZ: works for me 👍

zyte_spider_templates/middlewares.py
```diff
@@ -28,13 +29,17 @@ class CrawlingLogsMiddleware:
       the fingerprints logged in Scrapy Cloud's request data.
     """
 
+    # Deprecated in practice, but there is no good way to deprecate it, since
+    # class properties that also work for class instances are not a thing.
+    # https://stackoverflow.com/q/128573
     valid_page_types = [
```
kmike, line 35: I think it can be better to remove it, to avoid confusion.

kmike (approved):

PR #69, opened 2024-09-27
Custom attributes extraction

wRAR:


wRAR:
Looks like the test failure on `min` is because of Pydantic 1.x not producing additional metadata parts (that we don't use anyway) for `Json` fields.

wRAR:
> Pydantic 1.x not producing additional metadata parts

Actually it's Pydantic 2.0, (we require 2.0+ and that code was added in 2.1). We can change the test expectations to be different on different Pydantic versions, or we can bump the minimum Pydantic version (no idea if that makes a difference for anything except the test itself).

kmike:
I think bumping Pydantic 2.x version is fine. It seems we still should keep 1.8 support for now, right?

wRAR:
We don't have 1.8 support in this project, as we have `pydantic>=2` since the beginning.

kmike:
But isn't the fix in scrapy-spider-metadata, not here, which supports old pydantic?

kmike (approved):

PR #74, opened 2024-10-18
Add a non-whitespace pattern to SearchQueriesParam.

wRAR:


Gallaecio:
Wouldn’t this prevent multi-word queries, like searching for “foo bar”?

wRAR:
Hmm, apparently it tries to match the entire value even without ^$

tests/test_serp.py
```diff
@@ -244,6 +244,7 @@ def test_metadata():
                         {"type": "null"},
                     ],
                     "description": "Input 1 search query per line (e.g. foo bar).",
+                    "pattern": ".*\\S+.*",
```
Gallaecio, line 247: > To include newlines use the (.|\r?\n)* pattern which avoids the use of regex flags/modifiers and has good support across regex libraries. ([src](https://json-schema.org/understanding-json-schema/reference/regular_expressions))

zyte_spider_templates/spiders/serp.py
```diff
@@ -18,6 +18,7 @@ class SearchQueriesParam(BaseModel):
         description="Input 1 search query per line (e.g. foo bar).",
         json_schema_extra={
             "widget": "textarea",
+            "pattern": r"(.|\r?\n)*\S+(.|\r?\n)*",
```
Gallaecio, line 21: > When defining the regular expressions, it's important to note that the string is considered valid if the expression matches anywhere within the string.
> https://json-schema.org/understanding-json-schema/reference/string#regexp

So, I’m thinking this should do the trick:

```suggestion
            "pattern": r"(.|\r?\n)*?\S",
```

Gallaecio (approved):

zyte_spider_templates/spiders/serp.py
wRAR: Well, the initial one didn't work on a multi-word input (neither does `r"\S"`) so I don't know if the quote is true. Or maybe the frontend code uses it incorrectly.

Gallaecio: > Or maybe the frontend code uses it incorrectly.

Oh, right, the frontend is most likely not using some JSON schema library, and instead uses its own implementation, which might be assuming an initial `^` (like Python’s `re.match` vs `re.search`).

kmike: This pattern is not super straightforward - would it be possible to add a test for it? I.e. check that the validation passes with input A, and fails with input B?

Gallaecio: Does `\S` pass those tests?

wRAR: It does.

I don't think we can emulate the frontend validatoion logic here?

kmike (approved):

PR #75, opened 2024-10-25
SERP: use serial pagination

Gallaecio:


zyte_spider_templates/spiders/serp.py
```diff
@@ -117,12 +130,14 @@ def start_requests(self) -> Iterable[Request]:
...
-                        search_url, "start", str(start)
-                    )
-                yield self.get_start_request(search_url)
+            yield self.get_serp_request(search_url)
 
     def parse_serp(self, response) -> Iterable[Serp]:
-        yield Serp.from_dict(response.raw_api_response["serp"])
+        serp = Serp.from_dict(response.raw_api_response["serp"])
+
+        next_start = serp.pageNumber * self._results_per_page
```
kmike, line 138: I wonder if it'd be more reliable to pass the pageNumber around in the spider, instead of relying on extraction (which has a chance to fail)

Gallaecio: Somewhat related, I myself wonder if relying on the totalOrganicResults number is reliable as well, to be honest. I went that way to prioritize avoiding extra requests over reliability, but maybe that’s not a good idea.

I also considered counting organicResults, and only following when they are 10, but I have been told sometimes they could be lower than 10 in a regular page with follow up pages. Ignoring those issues, this approach would result in 1 extra request in cases where the last page has 10 results.

I also considered going to the next page if organicResults are 0. This has the advantage that pages with 1 or 2 fewer organic results will not cause pagination to stop, and the disadvantage that you reliably get 1 extra empty results page every time.

Any thoughts on this?

kmike: I think we can keep the current approach for totalOrganicResults for now, and change it if there are issues with it. It seems there are tradeoffs, and it's not clear what's the best approach for stopping in practice. The case for 0 results is handled in the current code, which is good.

zyte_spider_templates/spiders/serp.py
```diff
@@ -97,10 +101,25 @@ def update_settings(cls, settings: BaseSettings) -> None:
...
+        warn(
+            (
+                "zyte_spider_templates.spiders.serp.GoogleSearchSpider.get_start_request "
+                "is deprecated, use get_serp_request instead."
+            ),
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        page_number = int(int(url_query_parameter(url, "start", "0")) / 10 + 1)
+        return self.get_serp_request(url, page_number=page_number)
```
kmike, line 114: I think that's fine to drop it.

zyte_spider_templates/spiders/serp.py
```diff
@@ -117,12 +136,30 @@ def start_requests(self) -> Iterable[Request]:
...
+                    "(cb_kwargs) indicating the page being parsed (e.g. 1 for "
+                    "the 1st page), and will yield a request for the next "
+                    "page if the next page could have additional results, "
+                    "instead of expecting the source callback to take care of "
+                    "pagination by yielding requests for multiple pages in "
+                    "parallel."
+                ),
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
kmike, line 163: It could be fine to drop all the warnings as well :) It's nice to have the warnings, but it seems we don't have many users, and it complicates the code, so it may be better to just do a backwards-incompatible change for now.

zyte_spider_templates/spiders/serp.py
```diff
@@ -117,12 +121,14 @@ def start_requests(self) -> Iterable[Request]:
...
-                    search_url = add_or_replace_parameter(
-                        search_url, "start", str(start)
-                    )
-                yield self.get_start_request(search_url)
-
-    def parse_serp(self, response) -> Iterable[Serp]:
-        yield Serp.from_dict(response.raw_api_response["serp"])
+            yield self.get_serp_request(search_url, page_number=1)
+
+    def parse_serp(self, response, page_number) -> Iterable[Serp]:
```
kmike, line 126: typing seems to be off, because it can yield Requests, not only Serp items

kmike (approved):

PR #77, opened 2024-10-31
Implement search support

Gallaecio:
To do:
- [x] https://github.com/scrapinghub/Formasaurus/pull/3
- [x] https://github.com/scrapinghub/extruct/pull/237
- [x] https://github.com/zytedata/zyte-common-items/pull/112
- [x] https://github.com/zytedata/zyte-common-items/pull/113
- [ ] ~~Log an error when a search request cannot be built from a start URL.~~


docs/customization/pages.rst
```diff
@@ -141,3 +142,101 @@ To extract a new field for one or more websites:
...
+
+    @handle_urls("example.com")
+    class ExampleComSearchRequestTemplatePage(SearchRequestTemplatePage):
+        @field
+        def url(self):
+            return "https://example.com/search?q={{ keyword|quote_plus }}"
+
+
+Reusing the default implementation
+----------------------------------
```
kmike, line 172: I still prefer not to document it, at least here. It would allow us to change the implementation without breaking the documented ways to use the spider. What do you think about moving it to the module docstring?

zyte_spider_templates/pages/search_request_template.py
```diff
@@ -0,0 +1,326 @@
...
+
+        if not isinstance(self.response.response, BrowserResponse):
+            raise UseFallback(
+                f"Cannot build a search request template for "
+                f"{self.response.url}. Try enabling browser rendering by "
+                f'setting "browserHtml" to True in the '
+                f"ZYTE_API_PROVIDER_PARAMS setting dict."
+            )
+
+        raise UseFallback(
```
kmike, line 313: What's the idea behind using UseFallback exception here?

zyte_spider_templates/pages/search_request_template.py
```diff
@@ -0,0 +1,326 @@
...
+            f"search URL as input URL instead of using the search "
+            f"queries input field. You can also manually implement "
+            f"search support for a given website "
+            f"(https://zyte-common-items.readthedocs.io/en/latest/usage/request-templates.html#writing-a-request-template-page-object). "
+            f"You can also submit a request to add support for a specific "
+            f"website in the public GitHub issue tracker "
+            f"(https://github.com/zytedata/zyte-spider-templates/issues) "
+            f"or open a private support ticket if you are a Zyte customer "
+            f"(https://support.zyte.com/support/tickets/new), but there "
+            f"is no guarantee that we will implement support for it."
```
kmike, line 325: I'd drop it; not sure if we want to keep website-specific search templates in this repo (at least not now probably).

zyte_spider_templates/pages/search_request_template.py
Gallaecio: I did not put too much thought in it, just thinking maybe in the future there may be a Zyte API implementation, and if this PO is still used but cannot find a search request template, we default to asking Zyte API for it.

Shall we instead return a 0-probability item and raise a `CloseSpider` exception from the callback?

kmike: Hm, interesting. It seems returning an item with 0 probability makes sense; we don't really support UseFallback exception at the moment. But I'm not sure if you need to close the spider, as there could be more than 1 seed urls.

zyte_spider_templates/pages/search_request_template.py
```diff
@@ -0,0 +1,318 @@
...
+                    continue
+                return top_results[builder_id]
+
+        if not isinstance(self.response.response, BrowserResponse):
+            logger.error(
+                f"Cannot build a search request template for "
+                f"{self.response.url}. Try enabling browser rendering by "
+                f'setting "browserHtml" to True in the '
+                f"ZYTE_API_PROVIDER_PARAMS setting dict."
+            )
```
kmike, line 306: Is it a common error you've faced during the tests?

Gallaecio: With the current implementation, during tests, 3194 domains worked with browser, while only 3113 worked with HTTP, so I guess you could say it is a real issue, it maybe not that significant.

If you are thinking that maybe we don’t need a browser-specific issue, and we should just suggest people to write a page object if things fail, I am fine with that.

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -241,8 +256,35 @@ def get_start_request(self, url):
...
+                    url=url,
+                    callback=self.parse_search_request_template,
+                    meta=meta,
+                )
+        else:
+            for url in self.start_urls:
+                yield self.get_start_request(url)
+
+    def parse_search_request_template(
+        self, response: DummyResponse, search_request_template: SearchRequestTemplate
```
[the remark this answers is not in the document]
Gallaecio: Great catch.

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -241,8 +256,37 @@ def get_start_request(self, url):
...
-            yield self.get_start_request(url)
+        if self.args.search_queries:
+            for url in self.start_urls:
+                meta: Dict[str, Any] = {
+                    "crawling_logs": {"page_type": "searchRequestTemplate"},
+                }
+                if self.args.crawl_strategy == EcommerceCrawlStrategy.full:
+                    meta["page_params"] = {"full_domain": get_domain(url)}
+                if self.args.extract_from == ExtractFrom.browserHtml:
+                    meta["zyte_api_automap"] = {"browserHtml": True}
```
kmike, line 267: Do you know if it works properly with AnyResponse dependency? I.e. the switch happens, and also there is no extra raw http request when the browser is used?

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -241,8 +256,37 @@ def get_start_request(self, url):
...
-            yield self.get_start_request(url)
+        if self.args.search_queries:
+            for url in self.start_urls:
+                meta: Dict[str, Any] = {
+                    "crawling_logs": {"page_type": "searchRequestTemplate"},
+                }
+                if self.args.crawl_strategy == EcommerceCrawlStrategy.full:
+                    meta["page_params"] = {"full_domain": get_domain(url)}
+                if self.args.extract_from == ExtractFrom.browserHtml:
+                    meta["zyte_api_provider"] = {"browserHtml": True}
```
Gallaecio, line 267: @wRAR Tried `meta["inject"] = [BrowserResponse]` here but it did not seem to work. Should it work? (as in, do we want it to work as a way to force browser over HTTP for AnyResponse, regardless of whether or not it works like that at the moment)

Gallaecio: @wRAR @kmike My bad, I tested without also declaring DynamicDeps on the callback, I see that’s also required.

zyte_spider_templates/spiders/ecommerce.py
Gallaecio: It most likely did not before [switching this to `zyte_api_provider`](https://github.com/zytedata/zyte-spider-templates/pull/77/commits/9dd4f3828d13065e8caf0d43ad82e567682dd3e3), seems to work as expected [with DynamicDeps](https://github.com/zytedata/zyte-spider-templates/pull/77/commits/cdaea6442c62fcf50374749d2e941b199c187fdf).

zyte_spider_templates/pages/search_request_template.py
kmike: I don't know! For the UI users, the message could be confusing, as the solution is not to change ZYTE_API_PROVIDER_PARAMS (it's not clear how to do it, and I'm not even sure it'd work), but to change the spider extract_from argument. The easiest seems to be to drop this message :)

zyte_spider_templates/params.py
```diff
@@ -108,6 +108,43 @@ class MaxRequestsParam(BaseModel):
...
 
+class SearchQueriesParam(BaseModel):
+    search_queries: List[str] = Field(
+        title="Search Queries",
+        description=(
+            "Turn the input URLs into search requests for these queries. You "
+            "may specify a separate search query per line. If search request "
+            "building fails, you can instead pass search URLs as input start "
+            "URLs, or customize the AI spider project with a search request "
+            "template page object (check the docs)."
```
kmike, line 119: > Turn the input URLs into search requests for these queries.

This reads a bit cryptic to me :( 

I wonder if it's ok to explain it on a lower level, closer to the implementation: a list of search queries to submit using the search form found on the input URL.

kmike: > or customize the AI spider project with a search request template page object (check the docs).

While technically correct, this applies to everything - e.g. to all the crawl strategies we provide, and to the extraction as well. I'd leave it out from the description; we need to communicate it, but we need to do it for everything somehow, so probably we should do it separately.

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -241,8 +257,40 @@ def get_start_request(self, url):
...
     def start_requests(self) -> Iterable[Request]:
-        for url in self.start_urls:
-            yield self.get_start_request(url)
+        if self.args.search_queries:
+            for url in self.start_urls:
+                meta: Dict[str, Any] = {
+                    "crawling_logs": {"page_type": "searchRequestTemplate"},
+                }
+                if self.args.crawl_strategy == EcommerceCrawlStrategy.full:
+                    meta["page_params"] = {"full_domain": get_domain(url)}
```
kmike, line 266: I wonder what's the use case for this.

Is the idea that navigation model fails (misses some items, or misses pagination), and by picking "full" strategy we can follow almost any link on the target page, as a workaround? Do you know if it works in practice - won't you get products which are unrelated to the search query submitted - and if so, what's the point of using it?

zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -241,8 +257,40 @@ def get_start_request(self, url):
...
+        response: DummyResponse,
+        search_request_template: SearchRequestTemplate,
+        dynamic: DynamicDeps,
+    ) -> Iterable[Request]:
+        probability = search_request_template.get_probability()
+        if probability is not None and probability <= 0:
+            return
+        for query in self.args.search_queries:
+            yield search_request_template.request(query=query).to_scrapy(
+                callback=self.parse_navigation,
```
kmike, line 289: Hm, I wonder if we should disable following subCategories on the target pages. When checking how it works, did you have cases where it'd be helpful to follow subcategories? 

If they're not useful in practice, following them can be harmful, as it's possible that the ML model will extract something which is not subCategory - this would lead to extraction of items unrelated to the search queries.

So, the question is - what is more frequent - cases with subcategories on search result pages, or cases where on search result pages subcategories are detected, but they're wrong?

zyte_spider_templates/params.py
Gallaecio: 5763840

zyte_spider_templates/spiders/ecommerce.py
Gallaecio: I cannot think of a use case. I’ll make the strategy incompatible with search queries.

Gallaecio: 8389d7b

kmike (approved):

PR #78, opened 2024-11-06
Google Search: Allow extracting items from SERP results

Gallaecio:
To do:
- [x] https://github.com/zytedata/zyte-common-items/pull/111
- [x] https://github.com/zytedata/zyte-common-items/pull/119

### DIFF
--- a/zyte_spider_templates/spiders/serp.py
+++ b/zyte_spider_templates/spiders/serp.py
@@ -104,7 +104,7 @@ ITEM_TYPE_CLASSES = {
 
 class SerpItemTypeParam(BaseModel):
     item_type: Optional[SerpItemType] = Field(
-        title="Item type",
+        title="Follow and Extract",
         description=(
             "If specified, result URLs are followed to extract the specified "
             "item type. Spider output items will be of the specified item "
--- a/zyte_spider_templates/spiders/serp.py
+++ b/zyte_spider_templates/spiders/serp.py
@@ -110,6 +110,7 @@ class GoogleDomainParam(BaseModel):
 
 class GoogleSearchSpiderParams(
     MaxRequestsParam,
+    SerpItemTypeParam,
     SerpMaxPagesParam,
     SearchQueriesParam,
     GoogleDomainParam,
--- a/zyte_spider_templates/spiders/serp.py
+++ b/zyte_spider_templates/spiders/serp.py
@@ -7,9 +7,10 @@ from scrapy.settings import SETTINGS_PRIORITIES, BaseSettings
 from scrapy_poet import DummyResponse, DynamicDeps
 from scrapy_spider_metadata import Args
 from w3lib.url import add_or_replace_parameter
-from zyte_common_items import (  # TODO: Add ForumThread to zyte-common-items; ForumThread,
+from zyte_common_items import (
     Article,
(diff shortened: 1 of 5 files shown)


zyte_spider_templates/spiders/serp.py
```diff
@@ -48,6 +58,83 @@ class SerpMaxPagesParam(BaseModel):
...
+class SerpItemTypeParam(BaseModel):
+    item_type: SerpItemType = Field(
+        title="Item type",
+        description="Data type of the output items.",
+        default=SerpItemType.serp,
+        json_schema_extra={
+            "enumMeta": {
+                # TODO: Add a test to make sure this is in sync with the enum class above.
+                # TODO: Try automating the generation of this metadata from the enum type above.
+                SerpItemType.serp: {
```
kmike, line 120: Hm, that's an interesting idea. I was thinking that the default here would be None, which means "don't go to the results, output the links".

wRAR (approved):

zyte_spider_templates/spiders/serp.py
```diff
@@ -48,6 +59,61 @@ class SerpMaxPagesParam(BaseModel):
...
+    SerpItemType.forumThread: ForumThread,
+    SerpItemType.jobPosting: JobPosting,
+    SerpItemType.product: Product,
+    SerpItemType.productList: ProductList,
+}
+
+
+class SerpItemTypeParam(BaseModel):
+    item_type: Optional[SerpItemType] = Field(
+        title="Item type",
```
kmike, line 107: Hm, here, in the title, "item type" doesn't read very well, as you noticed :(

zyte_spider_templates/spiders/serp.py
```diff
@@ -48,6 +59,61 @@ class SerpMaxPagesParam(BaseModel):
...
+        description=(
+            "If specified, result URLs are followed to extract the specified "
```
kmike, line 109: I wonder if we should make the description Google-specific. For example, there are organic and sponsored results; if I get it right, that's organic results which are followed, not sponsored - we should be explicit about it (unless we consider it a bug that sponsored results are not followed - which I'm not sure is a bug).

kmike: Something like "Follow the links in the organic search results, and extract the selected data type from the target pages"?

zyte_spider_templates/spiders/serp.py
```diff
@@ -58,6 +124,7 @@ class GoogleDomainParam(BaseModel):
 
 class GoogleSearchSpiderParams(
     MaxRequestsParam,
+    SerpItemTypeParam,
```
kmike, line 127: Should we provide extract_from option, which applies only for the items extracted from the linked websites, not to search results extraction itself?

kmike: Doesn't need to be in the same PR though.
We also should add custom attributes extraction, probably also separately.

kmike:
The implementation looks good to me.

zyte_spider_templates/spiders/serp.py
Gallaecio: 953091b?

kmike (approved):

PR #79, opened 2024-11-12
Google Search: results per page

Gallaecio:
To do:
- [x] Test that `ge=1` works on Scrapy Cloud, as this is the first time I recall using that in an optional field, and the UI may not support that.

### DIFF
--- a/zyte_spider_templates/spiders/serp.py
+++ b/zyte_spider_templates/spiders/serp.py
@@ -49,13 +49,11 @@ class SerpMaxPagesParam(BaseModel):
 
 
 class SerpResultsPerPageParam(BaseModel):
-    results_per_page: int = Field(
+    results_per_page: Optional[int] = Field(
         title="Results Per Page",
         description="Maximum number of results per page.",
         ge=1,
-        # NOTE: This should match the actual Google Search default, because
-        # when the default value is passed, it is not set in URLs.
-        default=10,
+        default=None,
     )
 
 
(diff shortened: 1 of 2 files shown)


zyte_spider_templates/spiders/serp.py
```diff
@@ -48,6 +48,17 @@ class SerpMaxPagesParam(BaseModel):
...
 
 
+class SerpResultsPerPageParam(BaseModel):
+    results_per_page: int = Field(
+        title="Results Per Page",
+        description="Maximum number of results per page.",
+        ge=1,
+        # NOTE: This should match the actual Google Search default, because
+        # when the default value is passed, it is not set in URLs.
+        default=10,
```
kmike, line 58: What do you think about having None as a default?

kmike: Aha, the issue is that we need to know results per page in a spider ayways, to paginate

Gallaecio: Yes.

I started by having `None`, and indicating in the field description that if nothing is specified, 10 is assumed for pagination purposes. But it felt unnecessarily verbose.

But I do like having `None` to have a difference between not setting `num` and setting it. Currently forcing `num=10` is not possible.

Once serp results provide a nextPage link, `None` may be even more useful, if we assume Google search might change results per page at some point of in specific scenarios, and that serp will detect that.

Let’s switch to `Optional[int]` with `default=None`, and leave 10 hard-coded as pagination for `None` until we can remove the hard-coding with `serp.nextPage`.

kmike (approved):

PR #82, opened 2024-11-15
Job posting spider.

wRAR:


zyte_spider_templates/spiders/job_posting.py
```diff
@@ -0,0 +1,269 @@
...
+                )
+            else:
+                yield self.get_nextpage_request(navigation.nextPage)
+
+    def parse_job_posting(
+        self, response: DummyResponse, job_posting: JobPosting
+    ) -> Iterable[JobPosting]:
+        probability = job_posting.get_probability()
+
+        # TODO: convert to a configurable parameter later on after the launch
```
kmike, line 188: We have a middleware in zyte-common-items to drop items below certain probability - what's the advantage of doing it in the spider?

kmike: Hmm, interesting - so, we're dropping custom attributes if the product probability is low; the middleware in zyte-common-items won't do it.

wRAR: Yeah, I admit I just copied the e-commerce spider and trimmed it down.

We may want to do something differently but also we may want to extract common code, and I think we need to wait until this and articles exist in the same tree.

zyte_spider_templates/spiders/job_posting.py
```diff
@@ -0,0 +1,247 @@
...
+from ..utils import load_url_list
+
+if TYPE_CHECKING:
+    # typing.Self requires Python 3.11
+    from typing_extensions import Self
+
+
+@document_enum
+class JobPostingCrawlStrategy(str, Enum):
+    category: str = "category"
```
Gallaecio, line 41: Shouldn’t we call this `pagination`, in line with the e-commerce spider?

zyte_spider_templates/spiders/job_posting.py
```diff
@@ -0,0 +1,247 @@
...
+
+    .. seealso:: :ref:`job-posting`.
+    """
+
+    name = "job_posting"
+
+    metadata: Dict[str, Any] = {
+        **BaseSpider.metadata,
+        "title": "Job posting",
+        "description": "Template for spiders that extract job posting data from job websites.",
```
Gallaecio, line 107:
```suggestion
        "description": "Template for spiders that extract job posting data from websites.",
```

zyte.com is not a job website, but you can extract job posting data from its jobs page.

Gallaecio (approved):
Looks good to me, besides the room for code-sharing with the e-commerce spider, which I am not sure how to best address and I am not sure whether or not to consider it in scope here, i.e. maybe we should handle that after we have articles and this already merged.

zyte_spider_templates/spiders/job_posting.py
```diff
@@ -0,0 +1,250 @@
...
+                    ),
+                },
+            },
+        },
+    )
+
+
+class JobPostingSpiderParams(
+    CustomAttrsMethodParam,
+    CustomAttrsInputParam,
```
kmike, line 74: It seems custom attributes item is never returned

wRAR:
I was asked to not merge this yet.

kmike (approved):

PR #90, opened 2024-12-04
Basic ProductList extraction.

wRAR:


zyte_spider_templates/spiders/ecommerce.py
```diff
@@ -290,16 +330,30 @@ def parse_search_request_template(
...
+                    meta["inject"].append(self._custom_attrs_dep)
             yield search_request_template.request(query=query).to_scrapy(
                 callback=self.parse_navigation,
-                meta={
-                    "crawling_logs": {"page_type": "productNavigation"},
-                },
+                meta=meta,
             )
 
     def parse_navigation(
```
kmike, line 345: It seems you're still following product links in this callback, even when self.args.dont_follow_product_links is True.

kmike (approved):

PR #93, opened 2024-12-10
Follow and extract default: None → Enum.nothing

Gallaecio:
captura-4

### DIFF
--- a/tests/test_serp.py
+++ b/tests/test_serp.py
@@ -327,7 +327,7 @@ def test_metadata():
                     "title": "Results Per Page",
                 },
                 "item_type": {
-                    "default": "nothing",
+                    "default": "off",
                     "description": (
                         "If specified, follow organic search result links, "
                         "and extract the selected data type from the target "
@@ -335,7 +335,7 @@ def test_metadata():
                         "data type, not search engine results page items."
                     ),
                     "enum": [
-                        "nothing",
+                        "off",
                         "article",
                         "articleList",
                         "forumThread",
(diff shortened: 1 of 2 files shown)


tests/test_serp.py
```diff
@@ -327,15 +327,15 @@ def test_metadata():
...
-                    "default": None,
+                    "default": "nothing",
                     "description": (
                         "If specified, follow organic search result links, "
                         "and extract the selected data type from the target "
                         "pages. Spider output items will be of the specified "
                         "data type, not search engine results page items."
                     ),
                     "enum": [
+                        "nothing",
```
kmike, line 338: It seems someone could get confused - they read "follow and extract nothing", and think what the hell the spider is going to do if it's extracting nothig? I'm not sure what's the better alternative though. Any ideas?

Gallaecio: We could replace “nothing” with “only_extract_serp” or “only_extract”.

Alternatively, we could name the option only “Extract”, replace “nothing” with “serp”, and consider the fact that following is needed for other types an implementation detail.

kmike: > We could replace “nothing” with “only_extract_serp” or “only_extract”.

This may work, but it also doesn't communicatd that the "follow" step is not applied. "only extract" is also a bit confusing, because it's not "extract" from "follow and extract", but another extract :)

> Alternatively, we could name the option only “Extract”, replace “nothing” with “serp”, and consider the fact that following is needed for other types an implementation detail.

I think we want to make users aware of the crawling logic even more eventually, so I wouldn't consider it entirely an implementation detail.

What about "disabled"?

kmike (approved):

PR #92, opened 2024-12-06
Job postings + ProductList extraction

wRAR:
#82 + #90, on top of #91

PR #96, opened 2024-12-12
Do not let request-building exceptions break crawling

Gallaecio:
Created a context manager to capture and log exceptions, and used it in every spider when building a request object, so that any exception from request building does not prevent the rest of the callback code from running normally.

Built on top of https://github.com/zytedata/zyte-spider-templates/pull/92 because the mockserver makes testing this change really easy (most of those tests failed upon adding those mailto: links, and work after the changes here).

### DIFF
--- a/zyte_spider_templates/spiders/article.py
+++ b/zyte_spider_templates/spiders/article.py
@@ -286,7 +286,7 @@ class ArticleSpider(Args[ArticleSpiderParams], BaseSpider):
 
         for url in self.start_urls:
             meta = {"request_type": request_type}
-            with self._log_exception:
+            with self._log_request_exception:
                 yield self.get_parse_request(
                     ProbabilityRequest(
                         url=url,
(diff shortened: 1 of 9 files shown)


zyte_spider_templates/spiders/article.py
```diff
@@ -286,15 +286,16 @@ def start_requests(self) -> Iterable[scrapy.Request]:
...
-            yield self.get_parse_request(
-                ProbabilityRequest(
-                    url=url,
-                    name=f"[{request_type.value.name}]",
-                    metadata=ProbabilityMetadata(probability=probability),
-                ),
-                meta=meta,
-                is_feed=False,
-            )
+            with self._log_exception:
```
kmike, line 289: Can we limit the exceptions which are logged, i.e. catch only the specific ones - something like this?

```
with self._log_exception(ValueError):
```

https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse indicates that urlparse fails with ValueError on incorrect URLs. I'm not confident it's the only exception which can be raised, who knows :)

kmike (approved):

PR #100, opened 2024-12-13
Fix collection name validation

kmike:


kmike:
This doesn't enable UI validation for some reason (looks like a bug), but at least it fails in a spider, and also adds a suggestion to the user not to use such collection names.

kmike:
I can confirm that when the field is made required, validation starts to work.

PR #106, opened 2024-12-16
Slugify and truncate the default collection name

Gallaecio:


zyte_spider_templates/_incremental/manager.py
```diff
@@ -22,11 +23,19 @@
...
 COLLECTION_API_URL = "https://storage.scrapinghub.com/collections"
 
 THREAD_POOL_EXECUTOR = ThreadPoolExecutor(max_workers=10)
 
 
+def _get_collection_name(crawler: Crawler) -> str:
+    if name := crawler.settings.get("INCREMENTAL_CRAWL_COLLECTION_NAME"):
+        return name
+    name = get_spider_name(crawler).rstrip("_")[:_MAX_LENGTH] + INCREMENTAL_SUFFIX
+    return slugify(name, separator="_", lowercase=False, regex_pattern=r"[^a-zA-Z0-9_]")
```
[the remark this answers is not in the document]
Gallaecio: We could replace any non-valid character with `_`. It would be trivial and reliable. I like the å → a conversion of slugify and lower-level libraries like unidecode, but here it makes sense to prioritize future-proofing.

zyte_spider_templates/_incremental/manager.py
```diff
@@ -22,11 +23,19 @@
...
 COLLECTION_API_URL = "https://storage.scrapinghub.com/collections"
 
 THREAD_POOL_EXECUTOR = ThreadPoolExecutor(max_workers=10)
 
 
+def _get_collection_name(crawler: Crawler) -> str:
+    if name := crawler.settings.get("INCREMENTAL_CRAWL_COLLECTION_NAME"):
+        return name
+    name = get_spider_name(crawler).rstrip("_")[:_MAX_LENGTH] + INCREMENTAL_SUFFIX
+    return re.sub(r"[^a-zA-Z0-9_]", "_", name)
```
kmike, line 36: This one may be problematic for another reason - all spiders with unicode names of the same length would silently get the same collection name, and reuse the fingerprint DB unintentionally.

Gallaecio: If we do not care about human readability of the collection names, we could replace characters with their UTF-8 hexadecimal (e.g. å → C3A5) or some other Unicode character ID.

Otherwise, I think we are back to slugify or at least Unidecode, with the caveat that the results may change over time since we have no control over those libraries. We could pin them in `zyte-spider-templates-project`, but eventually some user may find this problematic.

We could also use the spider numeric ID, since collections are project-specific. We can extract it easily from the middle of the job ID.

kmike: I also wonder what are the restrictions on the spider names. The solution could also be to bring the formats closer together (e.g. add more validation for virtual spider names). 

> Otherwise, I think we are back to slugify or at least Unidecode

Hm, using text-unidecode could be pretty stable. Unlike unidecode or python-slugify, it didn't have a release in many-many years :)

Gallaecio: Oh, nice! And its maintainer seems familiar :)

PyExplorer: What do you think about just disabling the default name creation and require an explicit collection name?

Gallaecio: Good point. I think it is worth considering. While it is a slightly worse initial user experience, it is the most reliable approach, no need for us to worry about any long-term issues with the default name generation.

We could make it a backward-compatible change by making it required only in the UI with some JSON schema customization, and logging a deprecation error (because warnings are not visible in the UI easily) to encourage users to set it.

kmike: I'm still unsure about requiring a collection name - it's reliable, but it's an additional step. I wonder how often would user face issues in practice with the current approach (unidecode), and if it worths an additional step.

What do you think about logging a warning if the collection name is not set explicitly, and the resulting name is mangled? With a suggestion to set it explicitly?

Gallaecio: > What do you think about logging a warning if the collection name is not set explicitly, and the resulting name is mangled? With a suggestion to set it explicitly?

That should help, yes. Though with the current UI, I wonder how many people will notice the warning.

Gallaecio: https://github.com/zytedata/zyte-spider-templates/pull/106/commits/6168c0682c13a80de14348a52099b38aafd41659

PR #107, opened 2024-12-19
Fix _replace_builtin

Gallaecio:
`_replace_builtin` is meant to find a component, disable it, and replace it by another with the same priority value.

However, due to a variable typo I probably made myself, this function was failing to see built-in components (those in the `*_BASE` settings), so they were never disabled, causing the built-in offsite middleware to remain enabled, which in turn caused offsite request to be ignored even if allowed through meta (that only our custom middleware takes into account).

### DIFF
--- a/tests/test_addon.py
+++ b/tests/test_addon.py
@@ -72,7 +72,7 @@ def _test_setting_changes(initial_settings, expected_settings):
                 "DOWNLOADER_MIDDLEWARES": {
                     MaxRequestsPerSeedDownloaderMiddleware: 100,
                     BUILTIN_OFFSITE_MIDDLEWARE_IMPORT_PATH: None,
-                    AllowOffsiteMiddleware: 500,
+                    AllowOffsiteMiddleware: 50,
                 },
                 "SCHEDULER_DISK_QUEUE": "scrapy.squeues.PickleFifoDiskQueue",
                 "SCHEDULER_MEMORY_QUEUE": "scrapy.squeues.FifoMemoryQueue",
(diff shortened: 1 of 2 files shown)


wRAR (approved):

PyExplorer (approved):

Gallaecio:
PR description added.

tests/test_addon.py
```diff
@@ -67,8 +71,8 @@ def _test_setting_changes(initial_settings, expected_settings):
                 "CLOSESPIDER_TIMEOUT_NO_ITEM": 600,
                 "DOWNLOADER_MIDDLEWARES": {
                     MaxRequestsPerSeedDownloaderMiddleware: 100,
-                    OffsiteMiddleware: None,
-                    AllowOffsiteMiddleware: 500,
+                    BUILTIN_OFFSITE_MIDDLEWARE_IMPORT_PATH: None,
+                    AllowOffsiteMiddleware: 50,
```
Gallaecio, line 75: As far as I can see from the code, this change in order (the offsite middleware will now come before the max-requests-per-seed middleware) is OK, probably even for the best. I hope it does not come back to bite me…

PyExplorer: :) It's interesting as all tests for the addon passed - might be this case was not covered :(

Gallaecio: The add-on was injecting the built-in middleware manually (the lines removed) before calling `_replace_builtin`, that’s why it was not caught by the tests.

PR #109, opened 2024-12-24
Add custom attribute support to the article spider.

wRAR:


Gallaecio (approved):

kmike (approved):
The code looks good, but we should have more tests for this feature before merging & releasing it.

Gallaecio:
Would something basic in line with `test_crawl` from https://github.com/zytedata/zyte-spider-templates/pull/117/files do the job, or are you thinking of something more?

PR #120, opened 2025-03-06
Add tests for the job posting spider using fake-zyte-api.

wRAR:


zyte_spider_templates/utils.py
```diff
@@ -15,8 +15,12 @@
 _URL_PATTERN = r"^https?://[^:/\s]+(:\d{1,5})?(/[^\s]*)*(#[^\s]*)?$"
 
 
-def get_domain(url: str) -> str:
-    return re.sub(r"^www\d*\.", "", parse_url(url).netloc)
+def get_domain(url: str, include_port: bool = True) -> str:
```
wRAR, line 18: I don't really like this but this seems like the easiest way to fix both middlewares, as `AllowOffsiteMiddleware` explicitly doesn't support ports while `OffsiteRequestsPerSeedMiddleware` uses `url_is_from_any_domain()` which  compares the ports.

tests/test_job_posting.py
```diff
@@ -628,3 +641,139 @@ def track_item(item, response, spider):
...
+    crawler = get_crawler(settings=settings, spider_cls=JobPostingSpider)
+    items = []
+
+    def track_item(item, response, spider):
+        items.append(item)
+
+    crawler.signals.connect(track_item, signal=signals.item_scraped)
+    await deferred_to_future(
+        crawler.crawl(url=str(jobs_website.make_url("/jobs/4")), max_requests=1000)
+    )
```
kmike, line 662: It seems there is a lot of common code in these tests. Could you please check if it can be refactored to one or more helper functions? See e.g. https://github.com/scrapy-plugins/scrapy-splash/blob/72a8788212746b938e1e4d45aad56ff27857924a/tests/utils.py#L18 (it's an example; not sure if that'd be a right abstraction here).

It'd be good to make tests to look more like this:

```py
spider_args = dict(
    url=str(jobs_website.make_url("/jobs/4"), 
    max_requests=1000,
)
# Default settings are used, but there also should be an argument to change 
# them per-test.
items = await crawl(JobPostingSpider, spider_args=spider_args) 

assert len(items) == 109
assert len(set(item.url for item in items)) == len(items)
assert len(set(item.jobPostingId for item in items)) == len(items)
```

tox.ini
```diff
@@ -1,15 +1,16 @@
...
+envlist = min,py39,py310,py311,py312,py313,mypy,linters,twine
 
 [testenv]
 deps =
     pytest
     pytest-cov
     pytest-twisted
     freezegun
+    zyte-test-websites @ git+https://github.com/zytedata/zyte-test-websites@main
+    fake-zyte-api @ git+https://github.com/zytedata/fake-zyte-api@main
```
Gallaecio, line 11: Should we freeze these on a specific tag or commit before merging?

tests/conftest.py
```diff
@@ -7,3 +16,38 @@ def mockserver():
 
     with MockServer() as server:
         yield server
+
+
+# copied verbatim from pytest-aiohttp
```
Gallaecio, line 21: I take it this was necessary. Could you elaborate on the comment why?

Gallaecio (approved):

PR #125, opened 2025-03-25
Add tests using the ecommerce test website.

wRAR:


tests/test_ecommerce.py
```diff
@@ -27,7 +33,18 @@
...
 from . import get_crawler
 from .test_utils import URL_TO_DOMAIN
-from .utils import assertEqualSpiderMetadata, get_addons
+from .utils import assertEqualSpiderMetadata, crawl_fake_zyte_api, get_addons
+
+if TYPE_CHECKING:
+    from aiohttp.pytest_plugin import AiohttpServer
+
+
+@pytest_twisted.async_fixture(scope="module")
```
kmike, line 42: why not to use "session" scope?

kmike (approved):

tests/test_ecommerce.py
wRAR: The options are "function" and "module".

Gallaecio (approved):

PR #127, opened 2025-04-24
Add tests for the articles spider using fake-zyte-api.

wRAR:
These are basic tests, not sure if we can test heuristics in some direct way as the only strategy the spider has (except "direct item") is "full". 
We could add more tests for `HeuristicsArticleNavigationPage` as we already have some in https://github.com/zytedata/zyte-spider-templates/blob/main/tests/pages/test_article_navigation_heuristics.py , though I'm not sure yet if the test website is more useful for that than the existing hardcoded HTML.

### DIFF
--- a/tests/test_article.py
+++ b/tests/test_article.py
@@ -1,13 +1,20 @@
+from __future__ import annotations
+
 from typing import Tuple, Type, cast
 from unittest.mock import patch
 
 import pytest
+import pytest_twisted
 import requests
 import scrapy
+from aiohttp.pytest_plugin import AiohttpServer
+from itemadapter import ItemAdapter
 from pydantic import ValidationError
 from scrapy.statscollectors import StatsCollector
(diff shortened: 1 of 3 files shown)


Gallaecio (approved):

tests/test_article.py
```diff
@@ -608,3 +623,93 @@ def test_crawl():
     assert requests[1].url == "https://example.com/link_1"
     assert requests[2].url == "https://example.com/link_2"
     assert requests[3].url == "https://example.com/link_3"
+
+
+@deferred_f_from_coro_f
+async def test_extract_jobs(zyte_api_server, articles_website):
```
kmike, line 629: Tests in this module have "jobs" in their name; should they be renamed to mention "articles"?

kmike (approved):
+1 to merge after fixing the test names.

